PR #29, opened 2019-10-17
Try opsd comparison notebook

Bachibouzouk:
I comment on the changed files the individual issues I had


postprocessing/OEP_MaStR_OPSD_Comparison.ipynb
```diff
@@ -1656,9 +1656,7 @@
     "import os\n",
     "import geopandas as gpd\n",
     "from shapely.geometry import Point\n",
-    "import upsetplot\n",
-    "\n",
-    "from helpers import *"
```
Bachibouzouk: `helpers` was not used here (and bokeh returns a 1500 line output in the notebook...)

postprocessing/OEP_MaStR_OPSD_Comparison.ipynb
```diff
@@ -1707,7 +1705,7 @@
     "# load renewables\n",
     "df_renewables = pd.read_csv(\n",
     "    \"data/OPSD/renewable_power_plants_DE.csv\",\n",
-    "    sep=\";\", encoding=\"ANSI\", parse_dates=[\"commissioning_date\"]\n",
+    "    sep=\";\", encoding=\"latin\", parse_dates=[\"commissioning_date\"]\n",
```
Bachibouzouk: I could not use 'ANSI' as it is only for Windows (according to [python's doc](https://docs.python.org/3/library/codecs.html#text-encodings)). Can you try to see if `'latin'` works by you too?

BenPortner (approved):
Looks good to me! Tested it locally and everything ran fine.

PR #41, opened 2019-10-22
Feature/updates

Bachibouzouk:


soap_api/mastr_power_unit_download.py
```diff
@@ -73,8 +73,7 @@ def get_power_unit(start_from, limit=API_MAX_DEMANDS):
 
 def download_power_unit(
         power_unit_list_len=2363200,
-        limit=API_MAX_DEMANDS,
-        ofname=None
```
Bachibouzouk, line 96: Please keep the option to give the `ofname` directly as an argument, the tests need it and the user is always free **not** to use it

soap_api/mastr_power_unit_download.py
```diff
@@ -102,10 +99,7 @@ def download_power_unit(
     log.info('Download MaStR Power Unit')
     log.info(f'Number of expected power units: {power_unit_list_len}')
 
-    if ofname is None:
-        ofname = csv_see
```
Bachibouzouk, line 125: Here instead of `csv_see` you can have `fname_all_units`

soap_api/utils.py
```diff
@@ -14,8 +14,31 @@
 log = logging.getLogger(__name__)
 import csv
 
-csv_see_dummy = f'data/bnetza_mastr_{DATA_VERSION}'
-csv_see = f'data/bnetza_mastr_{DATA_VERSION}_power-unit.csv'
+""" dummy stump for other file names """
+fname_dummy = f'data/bnetza_mastr_{DATA_VERSION}'
```
Bachibouzouk: I would prefer `fname_template` for variable name here

ghost: ok! will change

PR #52, opened 2019-11-28
Release/v0.9.0

Bachibouzouk:


soap_api/mastr_wind_download.py
```diff
@@ -54,8 +54,6 @@ def setup_power_unit_wind():
     else:
```
Bachibouzouk, line 59: Theoretically these commits should have gone in a v0.9.1, in their own branch, merged into `dev` then create branch v0.9.1 from `dev` and merge it into master.

Bachibouzouk:
@solar-c - after discussion with @Ludee we decided that this branch should not contain any further commits except for the documentation update.

@Ludee - I see you committed directly on this branch instead of committing to `dev and merging `dev` into this branch... The difference is that now we can't really keep working on `dev` until we make merge this release into `master` and in `dev` because there are code commits which exists only on the release branch...

Ludee (approved):

PR #84, opened 2020-05-15
Fix/#83 soap api wrapper

gplssm:
# Things changed

- [x] Restructure package such that subpackages (like soap_api, postprocessing) are located in package root folder open_mastr
- [x] Rewrote user and token management for SOAP API. Now, it's using `~/.open-MaStR/config.ini` for the user name and stores the token in the keyring. Both is requested from the user on the first time
- [x] Introduced `MaStRAPI()`, a python wrapper around MaStR SOAP API, which handles credentials, binds SOAP queries, passes credentials automatically to the latter and serializes the return values (returning a dict)
- [x] Updated README: little restructuring and larger section on using open-MaStR including low-level download (aka. `MaStRAPI()`)
- [ ] ~~Use `MaStRAPI()` in actual download functions~~

# To be discussed

- By restructering the package, I broke PR #81 and branch `data-release/2.4.0` file references. Thus, if this branch gets merged first, these changes have a lot of conflicts, or vice-versa. Thus, we need to take care of branch management here.
- ~~Use `MaStRAPI()` in actual download functions~~ Skip this step because it would require a lot of changes. Instead rewrite download functions in such a manner that is uses features of `MaStRAPI()` and saves code (now, there are quite a lot of repetitions, I would like to maintain less code :smirk: )


gplssm:
Gets closed in favor of #85 which targets dev-v0.1.0

PR #99, opened 2020-09-02
Unified parallel download

### LINKED ISSUE #73: Add process status in download_parallel_power_unit
Download status is only indicated with a status bar. 
Print badge number in addition to _INFO failed downloads: n_

### LINKED ISSUE #79: Workshop on multiprocessing
Hi @Bachibouzouk 
I struggle with the code for the multiprocessing for the download.
More download functions need a speed up with stable retry options.
I would like to develop a template I can use in this repo.
A refresh on the test would be nice as well.

- [tests/test_parallel_download.py](https://github.com/OpenEnergyPlatform/open-MaStR/blob/master/tests/test_parallel_download.py)
- [download_parallel_power_unit()](https://github.com/OpenEnergyPlatform/open-MaStR/blob/master/soap_api/mastr_power_unit_download.py#L144-L323)
- [download_parallel_unit_solar()](https://github.com/OpenEnergyPlatform/open-MaStR/blob/master/soap_api/mastr_solar_download.py#L135-L187)
- [download_parallel_unit_solar_eeg()](https://github.com/OpenEnergyPlatform/open-MaStR/blob/master/soap_api/mastr_solar_download.py#L353-L404)

### LINKED ISSUE #93: Provide execution time blacklist
Program should implement a function that halts/throttles execution at certain times of the day to respect MaStR API limits

### LINKED ISSUE #94: Implement timeout for unsuccessful API calls
The time of the last successful API call should be stored in a variable. The program should exit if there has been no successful call for a certain amount of time to avoid a situation where a crashed server still receives a large amount of requests.

### LINKED ISSUE #95: Implement unified parallelization interface
For easier parallelization, a generic function should be implemented that allows for passing any `get_power_unit_xxx` function as well as a list of candidate IDs as parameter. The function should then execute these function calls in parallel and return a list of the results.

### LINKED ISSUE #96: Generation of list of elements for download
Unit download seems to currently only be based on list indices and does not take into account units which are already downloaded. This seems quite inconvenient when resuming download sessions.
https://github.com/OpenEnergyPlatform/open-MaStR/blob/ae87a9ed49b6b379714fe8d5f266e5634ea6f3fb/soap_api/mastr_wind_download.py#L150

I would suggest improving upon this by generating the list of items to download by simply removing the already downloaded items from the overall list. Consider the following example:
```
a = read_power_unit_solar(fname_power_unit_solar)['EinheitMastrNummer'] # all generators
b = read_unit_solar(fname_solar_unit)['EinheitMastrNummer'] # already downloaded generators
c = a[~a.isin(b)]
```
Series `c` now contains all elements that occur in the global list but were not yet individually downloaded. Is there any reason why the current implementation is preferred? @Ludee

jadeaffenjaeger:
This separates the parallel download of elements into a single module.
The call interface of the parallel download functions is changed so that a list of `EinheitMastrNummer` of the items to be downloaded needs to be passed directly.

Furthermore this implements additional checks for:
- the time of day against a pre-defined blacklist to suppress execution during certain times of the day
- tracking of the time of the last successful retrieval of an item and error if database is not responsive

closes #96 
closes #95 
closes #94 
closes #93

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ python-dateutil
 psycopg2-binary
 multiprocess
 zeep
+tqdm
--- a/soap_api/main.py
+++ b/soap_api/main.py
@@ -18,6 +18,9 @@ __author__ = "Ludee; christian-rli"
 __issue__ = "https://github.com/OpenEnergyPlatform/examples/issues/52"
 __version__ = "v0.9.0"
 
+import datetime
+import time
+
 from soap_api.config import setup_logger
 from soap_api.mastr_general_download import get_mastr_time, get_mastr_time_auth, get_daily_contingent
 # from soap_api.mastr_power_unit_download import download_parallel_power_unit, download_power_unit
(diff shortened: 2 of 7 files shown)


Ludee (approved):
Merging fork to run test in feature branch.

Issue #202, opened 2021-10-27
Implement a XML parser to process the Gesamtdatenauszug


Ludee:
As discussed in #201, @henhuy developed a functional XML parser that converts the entire download into CSV files.
We will publish the code in this repo under the existing open license and invite anybody to improve the workflow.

FlorianK13:
I wrote a [parser](https://github.com/FlorianK13/python-mastrsql) that converts the XML files to a PostgreSQL database, but maybe this is a very specific use case and a feature that converts to CSV is more general.

Ludee:
I think these are both interesting use-cases. Since the data is quite large handling it in postgres also makes sense.
Should we merge the code and make all features available? I'm seeing that you already included merging the different files (e.g. solar 1-x) into one table. 
Small license issue: your code is MIT, this repo is AGPL.

FlorianK13:
I can change to AGPL as well, MIT was rather arbitrary. 
Since we need a running package on pypi within our project in the near future i will continue to implement the mastr package, but I will recommend to the open-mastr for people that look for more functionalities.

Ludee:
Having a stable release and creating a PyPI is the aim as well. I would be happy to join efforts to not have two packages doing the same. It is a little bit more effort now but will pay out in the future. 
I made you a member of this repo. Perhaps we schedule a short call to discuss the next steps and timeline?

FlorianK13:
"not have two packages doing the same" is an argument, of course. I think you already have my email, so you can propose a meeting and we can discuss the topic.

Ludee:
Open questions for the further developments
- One big package with all functions (API and XML)
- licenses AGPL or LGPL
- keep it simple (XML without docker possible)
- Where do we merge the XML-part
- Time scope: Release before Christmas

FlorianK13:
xml parser was implemented and will be released in 0.11.0

PR #225, opened 2022-04-26
Fix/soap api updated

chrwm:


chrwm:
All commits should be included in brach `feature-#202-xmlparser`

PR #251, opened 2022-05-20
Merge Hotfix from production to development

FlorianK13:
Need to have the same commits in development.


FlorianK13:
Tests now succeed, it seems like the error was due to the availability of marktstammdatenregister.de

Ludee (approved):

PR #262, opened 2022-06-03
Add warning message if the bulk download speed falls below a certain …

### LINKED ISSUE #256: Document download limitation from marktstammdatenregister.de
I have experienced now for a few times that the server from marktstammdatenregister.de sets a download speed limitation. At those times the bulk download works only with ~10kB/sec and needs several hours. Should we monitor the speed and give out a warning to the user if the speed is too slow.

"Warning: Currently the servers from marktstammdatenregister have restricted the download speed. Please try it again another time:"

deniztepe:
…limit #256

* The limit is set to 100kB/s
* The unit of tqdm bar is set to MB, since each iteration corresponds to one megabyte

closes #256

### DIFF
--- a/open_mastr/xml_download/utils_download_bulk.py
+++ b/open_mastr/xml_download/utils_download_bulk.py
@@ -35,13 +35,17 @@ def download_xml_Mastr(save_path: str) -> None:
         "Download has started, this can take several minutes."
         "The download bar is only a rough estimate."
     )
+    warning_message = (
+        "Warning: The servers from MaStR restricts the download speed."
+        " You may want to download it another time."
+    )
     print(print_message)
     url = get_url_from_Mastr_website()
     time_a = time.perf_counter()
     r = requests.get(url, stream=True)
     total_length = int(10000 * 1024 * 1024)
     with open(save_path, "wb") as zfile, tqdm(
-        desc=save_path, total=(total_length / 1024 / 1024)
+        desc=save_path, total=(total_length / 1024 / 1024), unit="MB"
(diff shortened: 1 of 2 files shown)


deniztepe:
The warning message should work fine now. Testing it is not intuitive, because you have to limit the download speed manually. Instead you can set the warning limit higher to something like 12 MB/s in `open_mastr/xml_download/utils_download_bulk.py:58` and run the following script:

```
from open_mastr.xml_download.utils_download_bulk import download_xml_Mastr
import os

# xml download
save_path = os.path.join(
            os.path.expanduser("~"), ".open-MaStR", "data", "Gesamtexport_test"
        )
download_xml_Mastr(save_path=save_path)
```

open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -35,13 +35,17 @@ def download_xml_Mastr(save_path: str) -> None:
         "Download has started, this can take several minutes."
         "The download bar is only a rough estimate."
     )
+    warning_message = (
+        "Warning: The servers from MaStR restricts the download speed."
```
FlorianK13: servers restrict (no s)

FlorianK13 (changes requested):
Only a small spelling mistake.

open_mastr/xml_download/utils_download_bulk.py
deniztepe: done

FlorianK13 (approved):
Looks good.

Issue #287, opened 2022-07-08
Adapt CI/CD for merge on production


FlorianK13:
A merge on production is usually done before a release on pypi.org. We should add the following workflow:

If pull request to production:

1. Build dist file with `python setup.py sdist`
2. Install the package with `pip install dist/open_mastr-0.11.4.tar.gz`
3. Run tests with the package from the distributable

deniztepe:
Merged with #295

PR #295, opened 2022-07-14
Separate ci workflows for develop and production branch #287

deniztepe:
* In case of a pull request on production branch the package is built by sdist and then the zipped artefact is pip installed instead of the directory with setup.py
This helps to verify that the built package is indeed self-sufficient.
* The test cases are reduced to two python versions but the macos is included as third os.

### DIFF
--- a/.github/workflows/ci-develop.yml
+++ b/.github/workflows/ci-develop.yml
@@ -1,0 +1,42 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches:
+      - develop
+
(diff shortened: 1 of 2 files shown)


FlorianK13 (approved):

FlorianK13:
Very nice. You have tested it on your forked branch right @deniztepe?

deniztepe:
Yes, but let me test briefly with a pr on production here as well

deniztepe:
Ignoring the repeating url error, the workflow seems fine. I am merging it into develop.

Issue #300, opened 2022-07-20
Permit data is only linked to one unit


FlorianK13:
When calling the Api method `GetEinheitGenehmigung` for a specific genMastrNummer, this usually returns a list of linked units (since usually an application includes a group of eg wind turbines and not only one).  Right now, only one unit is then saved to the table `permits`.

- [x] Write a string of comma seperated entries to the column when the method retrieve_additional_data(data_type="permit_data") is called
- [x] refactor the flatten_dict method in download.py
- [x] write unit tests for the refactored flatten_dict methods

FlorianK13:
Closed by #302

PR #302, opened 2022-07-21
Bugfix 300 linked permit units

FlorianK13:
Units that are linked to one permit are now all written to the permit table in a comma seperated string.
Issue: #300

### DIFF
--- a/open_mastr/soap_api/download.py
+++ b/open_mastr/soap_api/download.py
@@ -243,7 +243,6 @@ def flatten_dict(data, serialize_with_json=False):
     Flattens MaStR data dictionary to depth of one
 
     Parameters
-    Parameters
     ----------
     data : list of dict
         Data returned from MaStR-API query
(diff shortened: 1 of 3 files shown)


chrwm:
Thank you. I'll review and test in the afternoon.

open_mastr/soap_api/download.py
```diff
@@ -272,38 +275,87 @@ def flatten_dict(data, serialize_with_json=False):
...
+        if k in dic:
+            dic[k] = dic[k][v]
+
+    return dic
+
+
+def replace_linked_units_with_unit_identifier(dic: dict) -> dict:
+    """If data point in 'dic' has one or more VerknuepfteEinheit or
+    VerknuepfteEinheiten in its respective dict, then take first MaStRNummer
+    and insert it into key VerknuepfteEinheiten (flatten dict).
```
chrwm: This was the explanation I wrote for `dic[k] = dic[k][0][v]`. Since you changed it to you your described functionality of inserting a comma separated string, I would suggest:

```suggestion
    """If data point in 'dic' has one or more VerknuepfteEinheit or
    VerknuepfteEinheiten in its respective dict, the related units (VerknuepfteEinheiten) are inserted as comma-separated strings.
```

chrwm (approved):
Besides my suggested docstring change, looks good!
Also, testing the function yielded the expected results.

open_mastr/soap_api/download.py
FlorianK13: Thanks for correcting

PR #303, opened 2022-07-22
Feature 301 api locations

FlorianK13:
The API request for locations works again and linked unit identifiers (MastrNr) are written to the database instead of linked units.


chrwm (approved):
Looks good!

PR #306, opened 2022-07-25
Feature 269 export csv in chunks

### LINKED ISSUE #269: API export for solar fails due to low memory
When exporting the tables with the soap api, sometimes the export runs low on memory and fails.
Adapt the export function to avoid this.

chrwm:
Closes #269 

The feature will allow to export csv's in real chunks and hence enable csv-export on machines with little memory.


chrwm:
I'll address the refactoring of the csv export functions in a different issue.
Feel free to merge when you approve.

open_mastr/soap_api/download.py
```diff
@@ -395,7 +395,22 @@ def to_csv(df, technology):
 
     csv_file = os.path.join(data_path, filenames["raw"][technology]["joined"])
 
-    df.to_csv(csv_file, index=True, index_label="EinheitMastrNummer", encoding="utf-8")
+    if os.path.exists(csv_file):
+        df.to_csv(
+            csv_file,
+            mode="a",
```
deniztepe: The new method with 'append' leads to error, if the user tries to export the same tech for the second time, since the table already exists. The csv tables must be first deleted. Is the use case in this manner is intended?

deniztepe (approved):

open_mastr/soap_api/download.py
chrwm: I have not considered this to be a use case. In my view, the deletion of tables is a good way to avoid having exported csv tables from different database versions.

PR #310, opened 2022-07-26
Feature 293 api special rights

FlorianK13:
Corresponding Issue: #293


FlorianK13:
I still think this would be a nice feature, but we cannot really test it due to limited API rights. I'll close the issue for now, but it can be reopend if necessary.

PR #312, opened 2022-08-05
Feature 291 add black formatting as ci

### LINKED ISSUE #291: Add black formatting as CI
Should we add a black formatting github action like described in this video (at 6:36) https://www.youtube.com/watch?v=j1MbEYhYj_Y

deniztepe:
The pre-commit hook is activated, after pip installing the `pre-commit` package and running `pre-commit install` on console. You can test it on the feature branch, if you want to. I also added a short documentation of the feature.


FlorianK13:
I have adapted the flake8 file to be black compatible. There is still the pylint file (.pylintrc) in the main directory which is also for code formatting. Can we delete it since we only use black + flake8 ?

deniztepe:
I would say, yes?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
+    hooks:
+      - id: black
+        language_version: python3.9
```
FlorianK13, line 6: what does python3.9 mean here? Can it have conflicts with people using other versions?

deniztepe: It was on the default template. As far as I understood, it determines the python version for the hook. Black is called with this version. I assume it should be compatible with other versions such as 3.7...3-10.

FlorianK13: Let's merge it then and have a bugfix if it does not work.

FlorianK13 (approved):

Issue #325, opened 2022-08-17
Delete Code artefacts


FlorianK13:
There are a number of functions that are not used within the code. We should delete them to make development easier.

- [x] functions in `data_io.py` that were used in `cleaned_data`
- [x] `locations_to_csv` in `mirror.py`
- [x] `parallel.py` is not used
- [x] function `is_time_blacklisted` in soap_api/utils.py is only used in parallel.py (which is not used)
- [x] probably more if we search for them

FlorianK13:
The function cleaned_data from `data_io.py` is only used in postprocessing. But I think it does not work and it has no purpose anymore. As it says in the docstring:

Cleaning includes:

    * Removal of duplicates originating from migrated and directly entered data in MaStR
    which describes the same unit

Both sources of errors should not be present anymore since we have primary keys in the database which forbid duplicates.

FlorianK13:
`is_time_blacklisted` is only called in `parallel.py`. `parallel.py` is called nowhere in the code, but might be relevant for parallel download?

Either delete both `parallel.py` and method `is_time_blacklisted` or keep both.

deniztepe:
open_mastr.soap_api.mirror.MaStRMirror.create_additional_data_requests() function is not called anywhere. Is the explicit call of the function expected? Otherwise redundant function

FlorianK13:
While working on the documentation of the package, I now think that create_additional_data_requests is a function that can be called directly, so I uncommented it again within #434

PR #335, opened 2022-08-22
Feature #325 delete code artefacts

FlorianK13:
A few functions and modules need to be further discussed before final deletion.


chrwm (approved):
Thanks for cleaning the code! Looks fine!

PR #345, opened 2022-09-13
Feature 336 centralize hardcoded lists

deniztepe:
The script `constants.py` covers variables that are fix and needed throughout the module. This reduces the code length and improves the readability.


chrwm (approved):
Looks fine!

PR #370, opened 2022-11-11
Bugfix 367 postgres missing column handling

FlorianK13:


deniztepe:
I am testing it currently with a new download. It seems like the new columns could be arranged in orm.py . API and xml column names are slightly different. For instance:

```
Hoechstbemessungsleistung = Column(Float)
BiogasHoechstbemessungsleistung = Column(Float)
```

deniztepe:
Does the failing test mean there are new columns again?

`TypeError: 'eegAusschreibungZuschlag' is an invalid keyword argument for StorageEeg`

FlorianK13:
No, I deleted that column because it was not in the documentation. But it comes from the API download apparently, so we need to reintroduce it.

deniztepe (approved):

chrwm:
Does this mean there were new columns added to the XML? Or API? - in addtion to https://github.com/OpenEnergyPlatform/open-MaStR/issues/351? 
I don't understand the bug from the issue.

deniztepe:
New columns were added to XML, which were undergoing because of automatic column adding as error handling. This handling was not working with local postgres databases apparently.

PR #378, opened 2022-11-15
Release v0.12.1

### LINKED ISSUE #377: Release version v0.12.1
A few bugfixes as patch release:

- [x] Add versions to `changelog.md`
- [x] Fix and update package metadata in `setup.py`
- [x] Update `citation.cff`
- [x] Release on github

chrwm:


chrwm:
The current fail of the pypi tests workflow is due to the version number, which has to set to v0.12.1, we should do that in the last commit to the branch and only when we are sure that no other commits to the branch follow. This is because on test-PyPI version numbers can only be deployed once.

FlorianK13 (approved):

chrwm:
The final test version was successfully released on test-PyPI and all CI workflows were successfull.
I'll change the version number to the final release version number as the final and last commit on this branch before merging to production.

PR #389, opened 2022-11-24
Improve ci pipline

### LINKED ISSUE #382: Make version number dynamic in CI pipeline
For the [test-pypi-publish.yml](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/.github/workflows/test-pypi-publish.yml) we should make the version dynamic if possible. The idea is that it takes the version from `setup.py` and appends the current timestamp for example.

chrwm:
Closes #382


deniztepe (approved):

chrwm:
Increase of version number has been implemented in #407

PR #396, opened 2022-12-15
Add envvar section to docs

nesnoj:
This PR adds a short section on env variables to the docs.

Motivation: I'd like to have multiple DBs in parallel and setting the env var `SQLITE_DATABASE_PATH` seems to be the only way at the moment. I would like to find it in the docs.


nesnoj:
Shoot... `production` was set as default target :roll_eyes: 
Succeeded by #397

PR #399, opened 2022-12-15
Fix 395 db conn handling during db write errors (SQLA 2.0 compatibility)

### LINKED ISSUE #395: Import of bulk fails: `UNIQUE constraint failed: solar_eeg.EegMastrNummer`
Bulk import into DB fails with today's `Gesamtdatenexport_20221215.zip` (on `develop`, head: 85e9e3d4e4794d5f0e054f7d7dd38fb9f59171be):
```
[...]
File 'AnlagenEegSolar_19.xml' is parsed.
Data is cleansed.
File 'AnlagenEegSolar_20.xml' is parsed.
Data is cleansed.
File 'AnlagenEegSolar_21.xml' is parsed.
Data is cleansed.
  File ".../venv/lib/python3.8/site-packages/SQLAlchemy-2.0.0b4-py3.8-linux-x86_64.egg/sqlalchemy/engine/base.py", line 1933, in _exec_single_context
    self.dialect.do_executemany(
  File ".../venv/lib/python3.8/site-packages/SQLAlchemy-2.0.0b4-py3.8-linux-x86_64.egg/sqlalchemy/engine/default.py", line 743, in do_executemany
    cursor.executemany(statement, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: solar_eeg.EegMastrNummer
[...]
```
Apart from the expectation the BNetzA should have some basic integrity checks to ensure unique IDs, wouldn't the tool benefit from such a check before DB insert? Or do we have it and it simply doesn't cover this column?

nesnoj:
Fixes #395 

This fix allow to run open-mastr with SQLAlchemy v2.0 by explicitly creating a connection at some places (also works with <2.0).
It works out (tested with full run from scratch) but is not be understood as a general migration to SQLA v2.0. For the latter, the entire codebase should be re-checked (which needs quite some time I cannot afford).

As possible alternative, the version could be restricted to `<2.0` which will cause trouble in the long run but may be ok for now..

What do you think, how shall we proceed?

PS: I did not find a section "Unreleased" or sth. in the changelog to document this fix..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
 ### Changed
 - Repair the csv export [#384](https://github.com/OpenEnergyPlatform/open-MaStR/pull/384)
 - Replace numeric value in hydro_extended [#392](https://github.com/OpenEnergyPlatform/open-MaStR/pull/392)
+- Fix DB query errors when using SQLAlchemy v2.0 [#399](https://github.com/OpenEnergyPlatform/open-MaStR/pull/399)
 ### Removed
 
 ## [v0.12.1] Patch release - 2022-11-15
--- a/open_mastr/xml_download/utils_write_to_database.py
+++ b/open_mastr/xml_download/utils_write_to_database.py
@@ -278,9 +278,11 @@ def write_single_entries_until_not_unique_comes_up(
     table = tablename_mapping[xml_tablename]["__class__"].__table__
     primary_key = next(c for c in table.columns if c.primary_key)
 
-    key_list = (
-        pd.read_sql(sql=select(primary_key), con=engine).values.squeeze().tolist()
-    )
+    with engine.connect() as con:
+        with con.begin():
+            key_list = (
+                pd.read_sql(sql=select(primary_key), con=con).values.squeeze().tolist()
+            )
     df = df.set_index(primary_key.name)
     len_df_before = len(df)
     df = df.drop(labels=key_list, errors="ignore")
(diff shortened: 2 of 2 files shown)


chrwm:
> PS: I did not find a section "Unreleased" or sth. in the changelog to document this fix..

When reading this I hope you did not branch off from `production`, which is the default branch in this repo unlinke in other OEF repositories. If you switch to `develop` you'll find the unrelease section. We might consider switching that with regards to the [supported keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

nesnoj:
> > PS: I did not find a section "Unreleased" or sth. in the changelog to document this fix..
> 
> When reading this I hope you did not branch off from `production`, which is the default branch in this repo unlinke in other OEF repositories. If you switch to `develop` you'll find the unrelease section. 

No, I started from `develop`. But I guess I checked the changelog when I was on `production`....here comes the entry

deniztepe (approved):
Thank you for your finding out the issue and your quick fix. As you said the `engine.execute()` option does not work from SQLA 2.0 on. Since the v1.4 is still the latest stable release on pypi, no one has probably realized the error. But it is great that we have a flexible fix in advance. I would exclude the other option with the version number restriction. 

I will approve the PR and create a new issue for a complete check of the source code for the SQLA 2.0 compatibility.

nesnoj:
Thanks @deniztepe for your toughts and the issue!

Before I merge this, I'll perform some further tests on functionality.

PS: I was wondering how a beta version can make it into the package base, usually the latest available, stable version is used. And, for some reason, they used `Development Status :: 5 - Production/Stable` for the beta... :thinking:

nesnoj:
After testing other features I found more lines that do not work properly, e.g. the CSV export [here](https://github.com/OpenEnergyPlatform/open-MaStR/blob/987f214a71760d3f01e928da1354826638a349f1/open_mastr/soap_api/mirror.py#L1226-L1231).
Therefore I'll close this PR as it carries the risk of unintended behavior or errors in modules which have not been tested explicitly.

However, the **branch should be kept for future work**, I'll document it in #403 and #404.

Issue #386, opened 2022-11-21
Update automatic metadata creation


chrwm:
Update to newest OEMetadata standard and update hardcoded information.

Merge [create.py](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/open_mastr/soap_api/metadata/create.py) and [mastr_datapackage](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/open_mastr/soap_api/metadata/mastr_datapackage.json)

open tasks:

- [x] merge current dev into branch
- [ ] update to [oemetadata v15 ](https://github.com/OpenEnergyPlatform/oemetadata/blob/develop/metadata/v152/template.json)
- [ ] update hardcoded fields in datapackag_base
- [ ] include bnetza example information for each column at the end of `description`
- [ ] add additional columns from open-mastr in each resource
- [ ] create metadata for additional tables
- [ ] set zenodo URI for each resource
- [ ] register in databus for first time and set permanent URI
- [ ] refine ontological annotation

areleu:
Where are you planning on getting the updated metadata. I figured the data model is all across the place. Most of the info can be found in their web-sever documentation: https://www.marktstammdatenregister.de/MaStRHilfe/subpages/webdienst.html

I tried to parse the html version of the tables: https://www.marktstammdatenregister.de/MaStRHilfe/files/webdienst/Funktionen_MaStR_Webdienste_V24.2.139.html 

But matching it to the database needs some work, so before I do anything I wanted to know what has been done so far. 

Otherwise, the text descriptions are only to be found in the PDF documentation, AFAIK. 

Or is the plan to have the fields of the data package always hard coded?

PR #401, opened 2022-12-15
Refactor csv export

### LINKED ISSUE #400: Refactor csv export
The csv export is scattered over multiple files, classes and methods, which makes it hard to maintain.

- [x] remove statistikFlag, depreciated with release [Version: 1.2.73](https://www.marktstammdatenregister.de/MaStRHilfe/subpages/releasenotes.html)
- [x] Refactor all csv-export related functions to helpers.py
- [x] Check refactoring yields same csv output
- [x] Ensure dataversion-XXXX-XX-XX folder is automatically created
- [x] Add table naming for additional tables
- [x] Improve code readability
- [x] Update docstrings
- [x] Clean imports
- [x] Check open FIXME
- [x] Update tests

chrwm:
Closes #400


chrwm:
A short summary of the most important changes:
3 main functions exist:

1. csv-export is triggered via `MaStR.to_csv()`: defines and collects tables to export
2. `open_mastr.utils.helpers.create_db_query`: creates db-queries for TECHNOLOGIES and ADDITIONAL_TABLES
3. `open_mastr.utils.helpers.db_query_to_csv`: exports db-query to csv for TECHNOLOGIES and ADDITIONAL_TABLES

- functions relevant to the csv_export are now located in helpers.py


**Pytest**
Only open_mastr.utils.helpers.db_query_to_csv is tested for:
```
The test checks for 2 random tech and 2 random additional tables:
1. If csv's have been created and encoded in 'utf-8' and are not empty
2. For techs, if limited (limit=60) EinheitMastrNummer in basic_units are included in CSV file
3. For additional_tables, if csv is not empty
```
It is due to the Markstammdatenregister has unstable columns (columns are renamed, new columns are added, columns are deleted). This makes testing with static columns impractical in my opinion.
@FlorianK13 @deniztepe if you have ideas for better testing, please make suggestions.

chrwm:
Volkswagen was not necessary ;)

open_mastr/utils/config.py
```diff
@@ -222,6 +222,11 @@ def _filenames_generator():
             {loc: f"{prefix}_{loc}_raw.csv" for loc in API_LOCATION_TYPES}
         )
 
+        # Add filenames for additional tables
+        filenames["raw"].update({"additional_table":
```
deniztepe, line 226: Since I already have a filenames.yml file. It is not updated and this led to an error during csv export. Manually removing the filenames and run the code works, but could we indicate or in the best case automate the update of config files? A user would get the same error after updating the package, right?

open_mastr/utils/helpers.py
```diff
@@ -410,3 +429,329 @@ def data_to_include_tables(data: list, mapping: str = None) -> list:
...
+    metadata_file = os.path.join(data_path, filenames["metadata"])
+    unit_type_map_reversed = reverse_unit_type_map()
+
+    with session_scope(engine=engine) as session:
+
+        # check for latest db entry for exported technologies
+        mastr_technologies = [
+            unit_type_map_reversed[tech] for tech in data
+        ]
+        newest_date = (
```
deniztepe, line 589: When I run `mastr.to_csv(["balancing_area"] ` following error occurs
```
  File "C:\Users\deniz\Documents\open-MaStR\open_mastr\mastr_download_deniztest.py", line 59, in <module>
    mastr.to_csv(["balancing_area"])
  File "C:\Users\deniz\Documents\open-MaStR\open_mastr\mastr.py", line 368, in to_csv
    save_metadata(data=technologies_to_export, engine=self.engine)
  File "C:\Users\deniz\Documents\open-MaStR\open_mastr\utils\helpers.py", line 590, in save_metadata
    session.query(orm.BasicUnit.DatumLetzteAktualisierung)
TypeError: 'NoneType' object is not subscriptable
```
Because the basic_units table remains empty, when only additional tables are exported. Can we maybe add a try-except block wrapping this newest date and have a fallback date (maybe DatumDownload or dataversion date)?

tests/test_helpers.py
```diff
@@ -1,17 +1,37 @@
...
-import pytest
-import sys
-from datetime import datetime
-from open_mastr import Mastr
 
 
+# Check if db is empty
+_db_exists = False
+_db_path = os.path.join(expanduser("~"), ".open-MaStR", "data", "sqlite", "open-mastr.db") # FIXME: use path in tmpdir when implemented
+if os.path.isdir(_db_path):
```
deniztepe: Either if  should be replaced with `os.path.isfile(_db_path)` or below in the for loop the entry must be checked and the path should be the sqlite directory. 

Currently,` _db_exists` is never going to be True, right?

deniztepe (changes requested):
Overall it works well. I guess, the commented three points are small issues.

open_mastr/utils/config.py
chrwm: Yes, that's right. I thought about this but wanted to tackle it in another PR before the next release as this one is really full already.
A simple solution would be to remove the `if not` here https://github.com/OpenEnergyPlatform/open-MaStR/blob/46f51056636404e2c706ab9eb69bf164083f4d54/open_mastr/utils/config.py#L143-L151
However, this would lead to an update of the filenames.yml everytime the Mastr class is called and longer runtimes, as the function is called in `setup_project_home()` which is called in `__init__.py`.
Better solutions would check for e.g. the version or include refactoring. But both options should be handled in another issue IMO.

open_mastr/utils/helpers.py
chrwm: This is rather because `technologies_to_export` is emtpy in case of additional_tables and the query fails. The backfill is performed before in https://github.com/OpenEnergyPlatform/open-MaStR/blob/46f51056636404e2c706ab9eb69bf164083f4d54/open_mastr/mastr.py#L341-L344

However, thanks for spotting. This will be tackled in https://github.com/OpenEnergyPlatform/open-MaStR/issues/386 and is already listed as action point there. Until then, I comment `save_metadata` out.

tests/test_helpers.py
chrwm: True, I'd opt for changing the path, so that `os.scandir` can be used still.

open_mastr/utils/config.py
deniztepe: I created a new issue #410. Your suggestion makes sense for me.

open_mastr/utils/helpers.py
deniztepe: Resolved with https://github.com/OpenEnergyPlatform/open-MaStR/pull/401/commits/dc883995de0ec41e678ddccc67f644bc5d21da68 at the moment.

tests/test_helpers.py
deniztepe: I renamed one parameter not to overwrite _db_path in 092383c088eed89b97189600895b328c64f2c285. If it seems correct for you, you can resolve conversation.

chrwm:
@deniztepe you can approve, if I met your suggestions.

deniztepe (approved):

nesnoj:
Hey @chwrwm, I'd like to reference the package prior to the CSV refactoring. However, the latest version 0.12.1 does not include all recent fixes. 
It would be very nice if you could prepare an in-between-release (at 1ba8c7ce0b024ef7f65ba792e6a33021b89e5848) without the CSV stuff.
Is that possible for you? If not, I could take take of this..

nesnoj:
> Hey @chwrwm, I'd like to reference the package prior to the CSV refactoring. However, the latest version 0.12.1 does not include all recent fixes. It would be very nice if you could prepare an in-between-release without the CSV stuff. Is that possible for you? If not, I could take take of this..

Hmm, maybe it is not necessary. Did the CSV format change with the refactoring?

PR #402, opened 2022-12-15
Allow custom output path specification

### LINKED ISSUE #394: How to change project dir()
Hello everyone, 

for the past two months, I've been working on the Open-MaStR tool. It is being used for our project, Wasserstoffatlas (https://www.wasserstoffatlas.de/). This is a really helpful package for getting MaStR data, starting from the download of an XML file to exporting it as a CSV.

However, we have recently encountered a problem accessing the final CSV file [ bnetza_mastr_XXX_raw] from the home directory when given a path. Is it possible to change the project directory to CWD rather than the home directory so that we can access the file more easily?

We are setting everything up in such a way that it is implemented as an automated process for generic use on multiple devices, so if we have our file in CWD then it will be simpler for this to work.

Please help me with this issue @Ludee @FlorianK13 @deniztepe @chrwm 

Thank you

chrwm:
Closes #394 

Parameter currently named output_path because it will probably on specify the data folder with: xml-dump, csv file export,  database folder.
Logging and config folder are created before the Mastr class is called. Thus, user-inputs to Mastr are not forwarded.

More commits will be pushed once #401 is merged.


chrwm:
Should be tested once on Windows I suggest.

FlorianK13:
Is there the possibility to add a unit test for the output path?

chrwm:
I could unit test `get_output_dir()` and `get_data_version_dir()`. In its core the test would be if `os.environ.get()` and `os.path.join()` work as expected. That's the only thing the functions do. I assume that this will should have been tested in `os` package. 
I'm unsure about the added value of this test.

FlorianK13:
@chrwm The idea here is to have the full .open-mastr folder at the given path? Or only the csv files?

chrwm:
The latter, everything within `.open-mastr/data`

FlorianK13 (approved):

PR #426, opened 2023-02-06
Feature 418 delete docker

### LINKED ISSUE #418: docker-compose.yml does not exist in package
When you install the package using `pip install open-mastr` the docker-compose.yml file is not part of the package. Hence, `engine='docker-postgres'` does not work.

FlorianK13:


open_mastr/mastr.py
```diff
@@ -21,14 +21,13 @@
     data_to_include_tables,
     create_db_query,
     db_query_to_csv,
-    save_metadata,
```
chrwm, line 24: For now this import "save_metadata" can be deleted, but needs to be reintroduced when the function is fixed and operational again.

chrwm (approved):
Looks fine.

open_mastr/mastr.py
FlorianK13: Sorry, I deleted it since I tried to handle all the flake8 error messages.

PR #488, opened 2024-02-25
remove webscraping for URL

Johann150:
The URL format has been changed and no longer includes the random string. This means the web scraping with BeauitfulSoup is no longer necessary.

### PR-Assignee
~~📙 Update the documentation~~ no API change, not applicable

### Reviewer
- [x] 🐙 Follow the [Reviewer Guidelines](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md#40-let-someone-else-review-your-pr)
- [x] 🐙 Provided feedback and show sufficient appreciation for the work done

### DIFF
--- a/open_mastr/xml_download/utils_download_bulk.py
+++ b/open_mastr/xml_download/utils_download_bulk.py
@@ -5,7 +5,6 @@ from zipfile import BadZipfile, ZipFile
 
 import numpy as np
 import requests
-from bs4 import BeautifulSoup
 from tqdm import tqdm
 
 # setup logger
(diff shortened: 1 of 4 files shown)


FlorianK13:
Hi @Johann150, I'll have a look at your PR next week.

open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -5,30 +5,73 @@
...
+        release = 2
+    elif when.tm_mon > 10 or (when.tm_mon == 10 and when.tm_mday > 1):
+        release = 2
+
+    # only the last two digits of the year are used
+    year = str(year)[-2:]
+
+    return f'{year}.{release}'
+
+def gen_url(when: time.struct_time = time.localtime()) -> str:
```
FlorianK13, line 57: First of all thanks for this very useful PR @Johann150 
One remark here: I think it would be great to have a fallback url for the case that the current data is not online yet. This could happen, as you described, before 04:00.
Maybe we do the following: When the url is used to download the "Gesamtdatenexport" this is wrapped in a `try` - `except` block. If it fails, the url is changed to the url from one day before and the download is started again?

What do you think? And do you have time to make this change? Otherwise I can also do it.

FlorianK13 (changes requested):

open_mastr/xml_download/utils_download_bulk.py
Johann150: I have made the change. It was previously not checked at all what the download request status code was, so instead of writing a giant `try` block I thought it would be a better idea to "just" check the status code instead.

There is one potential situation that I'm not quite sure about, when thinking about this a bit more. Since the download can take a few minutes, if by coincidence you were to start the download right before the new file is published, I don't know if the rest of the old file will be correctly downloaded. But maybe that is a very hypothetical and contrived situation that does not really need to be considered.

FlorianK13: Yes, I think that won't happen too often and people can than rerun the download.

FlorianK13:
Tests never run successfully for PRs from outside as secret API credentials are not revealed

FlorianK13 (approved):

PR #598, opened 2025-01-12
[Feature #546]: Increase parsing speed by 47% by efficiently reading and writing the XML

### LINKED ISSUE #546: Increase parsing speed
This task contains several steps:

1. Search different ways that might increase parsing speed. Parsing is done right now by the `pandas.read_xml` method [here](https://github.com/OpenEnergyPlatform/open-MaStR/blob/d17a92c17912f7f96814d309859dc495f942c07a/open_mastr/xml_download/utils_write_to_database.py#L158). Several alternatives are:
  * `polars`, `duckdb`, `pyspark` might have xml parsers and might be faster
  * use plain xml parsing from python (without pandas)
  * ...

2. Writing to sqlite database right now is done by `pandas.to_sql` [here](https://github.com/OpenEnergyPlatform/open-MaStR/blob/d17a92c17912f7f96814d309859dc495f942c07a/open_mastr/xml_download/utils_write_to_database.py#L182C5-L182C26). There might be other faster methods depending on step 1.
3. Construct a benchmark in an own repository. Use a benchmark xml file from the Marktstammdatenregister and test different implementations for parsing them.
4. Decide for a best method and implement it in `open-mastr`

AlexandraImbrisca:
## Summary of the discussion

Improved the parsing speed by addressing 2 of the current bottlenecks:
- Reading the XML files into DataFrames
- Writing the DataFrames to the SQLite database

Over the last months I have analysed multiple alternatives and documented the results [here](https://docs.google.com/document/d/1ViGtGNq2wqvkYAGaH5um_3oJISxBvzuafAvo-q7WOuU/edit?tab=t.d7787cs8czgw) (please feel free to request access! or let me know if you prefer accessing the document in a different way).

Based on our MaStR dataset and its composition, the fastest parsing relies on:
- Using the lxml parser in combination with the pandas.read_xml function ([linked analysis](https://docs.google.com/document/d/1ViGtGNq2wqvkYAGaH5um_3oJISxBvzuafAvo-q7WOuU/edit?tab=t.d7787cs8czgw#bookmark=id.m874312my3q1))
- Bulk inserting the data using plain SQL statements ([linked analysis](https://docs.google.com/document/d/1ViGtGNq2wqvkYAGaH5um_3oJISxBvzuafAvo-q7WOuU/edit?tab=t.d7787cs8czgw#bookmark=id.4htp1a32kqlb))

By updating these 2 parts of the parsing logic, the execution time decreases by 47%. As a next step, I'll implement the parallelized version that reduces the overall time even more.

The current changes have been tested locally by using the [sqldiff](https://www.sqlite.org/sqldiff.html) to compare the resulting databases. I also added unit tests for all of the new and existing functions from the `utils_write_to_database`.

Fixes #546 

### Updated
- Updated the parsing logic [(#546)](https://github.com/OpenEnergyPlatform/open-MaStR/pull/546)

### PR-Assignee

### Reviewer


nesnoj:
Thanks a lot for this improvement @AlexandraImbrisca.
Will the parallelization be part of this PR or will you open a new one? In the latter case please assign a reviewer, thx!

AlexandraImbrisca:
I was thinking about creating a separate pull request to keep the changes easier to review. I unfortunately can't assign the reviewer myself (at least the UI doesn't allow me 🙈 )

FlorianK13:
Pytest is failing as usual for PRs from outside (This is a known error, where the API credentials saved in this github repo are not revealed to code coming from outside due to security reasons).

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -392,3 +293,51 @@ def find_nearest_brackets(xml_string: str, position: int) -> tuple[int, int]:
...
+    df = cast_date_columns_to_string(xml_table_name, df)
+
+    # Create SQL statement for bulk insert. ON CONFLICT DO NOTHING prevents duplicates.
+    insert_stmt = f"INSERT INTO {sql_table_name} ({','.join(column_list)}) VALUES ({','.join(['?' for _ in column_list])}) ON CONFLICT DO NOTHING"
+
+    for _ in range(10000):
+        try:
+            with engine.connect() as con:
+                with con.begin():
+                    con.connection.executemany(insert_stmt, df.to_numpy())
```
FlorianK13, line 454: One mistake that is probably my fault, since we never really talked about that: The code needs to work for two different database types: `sqlite` and `postgres`. It seems like postgres does not provide a `con.connection.executemany`. 
I assume that maybe 90% of users will have a sqlite database. Hence I see two possible solutions:
1. There is a way to change this code so that it works for postgres as well but does not decrease the parsing speed for sqlite.
2. You create two functions, one that is called when the engine is of type sqlite to use the fast code, and one that is called if the engine is of any other type. Then we use the old pandas.to_sql function and hope that pandas can handle all the other engine types.

FlorianK13: Do you agree @AlexandraImbrisca or do you see another way to go?

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -154,80 +149,26 @@ def correct_ordering_of_filelist(files_list: list) -> list:
...
-
-    # Add Column that refers to the source of the data
-    df["DatenQuelle"] = "bulk"
-    df["DatumDownload"] = bulk_download_date
-    return df
+def read_xml_file(f: ZipFile, file_name: str) -> pd.DataFrame:
+    """Read the xml file from the zip file and return it as a DataFrame."""
+    with f.open(file_name) as xml_file:
+        try:
+            return pd.read_xml(xml_file, encoding="UTF-16", parser='etree')
```
FlorianK13: We use black as a code formatter, so that code style is the same everywhere in the repository. Can you install the pre-commit hooks as described [here](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md#23-commit-your-changes) or directly install the code formatter [black](https://black.readthedocs.io/en/stable/). When running black on this file, it would do a few changes (like changing the 'etree' to "etree").

open_mastr/xml_download/utils_write_to_database.py
AlexandraImbrisca: Oh, thanks for mentioning! I completely missed that and I was under the impression that only sqlite databases are allowed. 

I'd prefer option 2 for now. I can bring back the old implementation for non-sqlite and use the new one only for sqlite. I will need to look into postgre and what is required to adjust the new approach, but hopefully that should decrease the time as well.

open_mastr/xml_download/utils_write_to_database.py
AlexandraImbrisca: Sure, I will do that! Thanks for catching these inconsistencies

open_mastr/xml_download/utils_write_to_database.py
AlexandraImbrisca: While I'm trying to set up PostgreSQL, is [this](https://github.com/OpenEnergyPlatform/open-MaStR/blob/bd44ce26986a0c41b25250951a4a9a391723d690/docs/advanced.md?plain=1#L25) the usage you are referring to?

FlorianK13: Yes, exactly. Let's stick with option 2 for now. I think getting this PR and the PR about parallelization merged has a higher priority right now.

AlexandraImbrisca: Sounds great, thank you! I updated the code to switch between the 2 implementations of the ```add_table_to_database``` function based on the dialect. The rest should remain the same since it doesn't rely on the database.

I parametrized the test for add_table_to_database and the functions produce the same output. Please take another look whenever you have time! I'll start the parallelization PR

FlorianK13:
I ran the following with the data from 15.01.
```python
db = Mastr()
db.download(date="existing", data="deleted_market_actors")
```
and got this error:
```
line 455, in add_table_to_sqlite_database: con.connection.executemany(insert_stmt, df.to_numpy())
sqlite3.IntegrityError: NOT NULL constraint failed: deleted_market_actors.MastrNummer
```

When using a postgres engine, this error does not appear. As far as I understand it, the error is catched by the pandas.to_sql function? At least it seems like neither the `delete_wrong_xml_entry` nor the `write_single_entries_until_not_unique_comes_up` is called. Can you maybe implement the following:

1. If engine=sqlite, use the fast sqlite method
2. If an error occurs, try to use the slower non_sqlite method also for sqlite databases for this one xml file

By this, the one xml file from deleted_market_actors would be slower to parse, but the `db.download` would not crash. What do you think @AlexandraImbrisca ?

AlexandraImbrisca:
Could you please try again? I previously removed the call to ```write_single_entries_until_not_unique_comes_up``` because I thought that "ON CONFLICT DO NOTHING" should handle this exception completely. I'm curious if the error is caught and handled correctly now by the SQL method

I agree that it would be great to switch to the non-SQL method in case of any other errors, so I added a second call to that function

FlorianK13:
Yes, now it does not crash anymore.

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -26,75 +27,81 @@ def write_mastr_xml_to_database(
...
-    # drop the content from table
-    orm_class.__table__.drop(engine, checkfirst=True)
-    # create table schema
-    orm_class.__table__.create(engine)
+def create_database_table(
+    engine: sqlalchemy.engine.Engine, xml_table_name: str
+) -> None:
+    """Create the table in the database if it does not exist."""
+    orm_class = tablename_mapping[xml_table_name]["__class__"]
+    if not inspect(engine).has_table(orm_class.__tablename__):
```
FlorianK13: I think I would keep the old process of:
1. deleting the old table
2. creating a new empty table
3. Filling this empty table with new data

Reason:
They can change units, eg change the value of `power` or so. If we do not overwrite old data, we will not have this change in our database. We could of course also have a combined Primary Key together with the data column indicating recent changes to the data, but I think this adds more complexity then necessary. I would argue for keeping this code:

```python
orm_class = tablename_mapping[xml_tablename]["__class__"]
orm_class.__table__.drop(engine, checkfirst=True)
orm_class.__table__.create(engine)
```

AlexandraImbrisca: Thanks a lot for explaining! I wasn't aware that this situation could happen. It makes a lot of sense then to recreate the table. I updated the implementation as you suggested

FlorianK13 (dismissed):

FlorianK13:
I would suggest the following:
- We can now merge this PR
- Afterwards we will look at the PR about parallelization
- When both are merged, I will also try the sqldiff stuff to check that the original package on pypi and the new code produce the same databases

Do you agree @nesnoj @AlexandraImbrisca ?

nesnoj:
> I would suggest the following:
> 
> * We can now merge this PR
> * Afterwards we will look at the PR about parallelization
> * When both are merged, I will also try the sqldiff stuff to check that the original package on pypi and the new code produce the same databases
> 
> Do you agree @nesnoj @AlexandraImbrisca ?

Thanks for checking and yes, we can handle it that way. But before the merge I'd like to have a quick look later today.

AlexandraImbrisca:
Sounds good to me! Please let me know if you have any other questions

nesnoj (dismissed):
Hey, I tested with sqlite and postgresql, both seem to work well (I did not check the data though).
(I only added a changelog entry)

I think we can get the postgreSQL inserts even faster. If you prefer we can do this in a separate PR, I'll add some notes to #546.

Thanks a lot for this improvement @AlexandraImbrisca!

PS: Could you please send me the analysis doc via mail? Thx!

AlexandraImbrisca:
Sorry for re-requesting the review again! I resolved a merge conflict which automatically dismissed your most recent approval.

@nesnoj Thanks a lot for testing and reviewing the changes! I completely agree that we can make the postgre faster. I'll do that in a separate PR and I'll send you the analysis doc over the email.

FlorianK13 (approved):

nesnoj (approved):

FlorianK13:
@AlexandraImbrisca this PR is finished, so I can merge it now, right?

AlexandraImbrisca:
Yes, thank you!

Issue #600, opened 2025-01-22
Speed up processing: DB insertion and multiprocessing


nesnoj:
Succeeding issue of recently closed #546.
I merged both topics into one issue as I think they interfere - e.g. choosing a specific DB insertion method might not work (efficiently) in multiprocessing. If you disagree, feel free to separate them.

1. Explore faster methods of writing to the database for
- sqlite
- postgres
2. Add multiprocessing for
- XML parsing
- Write to DB (if applicable due to concurrency -> table locks)

Notes on DB insertion: https://github.com/OpenEnergyPlatform/open-MaStR/issues/546#issuecomment-2603334204
Notes on parallelization: https://github.com/OpenEnergyPlatform/open-MaStR/issues/546#issuecomment-2604166231

Feel free to amend :)

FlorianK13:
we reached #600 🥇

AlexandraImbrisca:
Hi @nesnoj! Thanks for creating this issue. I'm testing my approach a bit more (different number of cores & different operating systems) and I'll create the PR. I've been developing and testing on MacOS & Linux and I'll continue with Windows. The approach is quite simple and uses the standard [concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html) library with a few options to optimize the access to the database

About writing the data to the postgre database: would you mind tackling this separately? I didn't get the chance to look into how optimize writing yet and the parallelization is database-agnostic right now

nesnoj:
> About writing the data to the postgre database: would you mind tackling this separately? I didn't get the chance to look into how optimize writing yet and the parallelization is database-agnostic right now

Sure, feel free to create a separate issue if that makes sense to you.

nesnoj:
Implemented, follow-up for postgres performance in #606

PR #601, opened 2025-01-26
[Feature #600]: Use multiprocessing to speed up the parsing

AlexandraImbrisca:
Follow up to [the previous PR](https://github.com/OpenEnergyPlatform/open-MaStR/pull/598):

- Using a [ProcessorPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) with 3 processors speeds up the execution time significantly 
- Depending on the operating system and technical specifications, we obtain a time decrease between 49.68% and 70.03% relatively to the previously optimized algorithm. In combination with the other improvements, this adds up to a 76,56% decrease from the initial, non-optimized implementation 
- Leveraging the standard multiprocessing functionality and carefully ordering the files leads to a safe optimisation across all tested environments


nesnoj:
Thank you @AlexandraImbrisca for the implementation and sending the detailed report which reads coherently!
Is this PR ready for review?

What I stumbled across so far:
- The CPU count is hard coded but should be configurable. Ideally via CLI but we do not have one, so maybe an environment variable could do the job. And I'm not sure whether >1 is an appropriate default, mp could also be promoted optionally. To be kept in mind: the base process uses 100 MB and each process about 1 GB. A standard office PC is equipped with ~8 GB so 3 processes might be ok. Alternatively, we could set a default of 1 and add a message like "Your system supports multiple CPU cores, you can increase the processing speed by setting env var ..."
What do you think @AlexandraImbrisca @FlorianK13 ?
- I tested with different CPU counts (Ryzen 7, Linux, SQlite DB, only `"solar"`). Concerning the processing speed increase my results are somewhat in line to yours:

| Cores | Time in s (SQLite) |
|-------|-----------|
| 3     | 394.8     |
| 4     | 316.6     |
| 5     | 292.4     |
| 6     | 280.8     |
| 8     | 279.5     |
| 10    | 280.6     |
| 12    | 276.4     |

The speed is stalling somewhere from 5 cores onward. I can imagine this drop in the speed increase is caused by a) the writing concurrency, b) other running processes on my laptop, c) number of parallel processes decrease once most of the tasks are done?
- Could you please explain why you chose parameters in `create_efficient_engine()` like this? And are they optimal for any number of CPUs?
- With 12 cores I occasionally(!) get the following error message:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) duplicate column name: InAnspruchGenommeneAckerflaeche
[SQL: ALTER TABLE solar_extended ADD "InAnspruchGenommeneAckerflaeche" VARCHAR NULL;]
```
(The column `InAnspruchGenommeneAckerflaeche` does not exist in our data model which isn't a problem - it is automatically added but there seems to be an issue with that in the mp)
- PostgreSQL is crashing [here](https://github.com/AlexandraImbrisca/open-MaStR/blob/436bb5406a81f327eefff4cc8a579c50f62d7d92/open_mastr/xml_download/utils_write_to_database.py#L86): `sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) invalid dsn: invalid connection option "timeout"`. The arg `timeout` in `connect_args` seems not to be understood by Postgres.
- The docs need to be updated
- Changelog entry is missing

AlexandraImbrisca:
Thanks a lot for the detailed review and suggestions @nesnoj!
- Default number of processes: I like the suggestion of keeping only 1 and adding a note! Since we are just introducing this feature, it might be helpful to make people aware of it and ask them to report any potential issues. How about we add that explanatory message and a link to the issues page to report any possible bugs / negative experiences? 
- Thanks for testing! From what I have read the general suggestion is to use CPU_count - 1 so I totally agree that increasing the number of cores relatively to the system makes sense. I think the performance stalls because of sqlite ([sqlite is not designed for write concurrency](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#database-locking-behavior-concurrency)) :( 
- The choice of parameters: sure, I'll leave some comments!
- Error when using 12 cores: oh, interestingly! OOC, is the exception caught or does the program terminate?
- Postgresql: I unfortunately tested mostly on sqlite:( I'll find a solution for this bug and test a bit more on postgresql
- Docs & changelog updates: sure thing! I'll create another commit for these updates

nesnoj:
Hey @AlexandraImbrisca !

> - Default number of processes: I like the suggestion of keeping only 1 and adding a note! Since we are just introducing this feature, it might be helpful to make people aware of it and ask them to report any potential issues. How about we add that explanatory message and a link to the issues page to report any possible bugs / negative experiences?

Sounds good to me.
What do you think @FlorianK13 ?

> I think the performance stalls because of sqlite ([sqlite is not designed for write concurrency](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#database-locking-behavior-concurrency)) :( 

An alternative way could be to create separate SQLite DBs and finally merge them. Dunno if this is a viable option..

> - Error when using 12 cores: oh, interestingly! OOC, is the exception caught or does the program terminate?

It terminates :(

FlorianK13:
> * Default number of processes: I like the suggestion of keeping only 1 and adding a note! Since we are just introducing this feature, it might be helpful to make people aware of it and ask them to report any potential issues. How about we add that explanatory message and a link to the issues page to report any possible bugs / negative experiences?

Sounds good to me as well!

AlexandraImbrisca:
Awesome, thanks a lot both! A few updates from my side:

- I introduced 2 new environment variables: one for using the recommended number of processes, one to set up a custom number of processes. I think that 2 variables are necessary since people might not be aware of what number of processes would perform the best, but it would be nice to allow them to customize it
- @nesnoj I think the "duplicate column name" exception occurs because of a race condition (i.e., 2 processes trying to add the same column at the same time). Please correct me if I'm wrong, but I think we can safely ignore this error since once we have introduced the missing columns, we reached our purpose 🤔 I added some more error handling. Could you please let me know if you are still able to reproduce this issue?
-  I fixed the PostgreSQL issue and generally tested more for PostgreSQL
- I updated the documentation and added a message to promote this feature

About merging the DBs: that might work, but it might get quite messy with many processes (i.e., we could end up with 10+ temporary DBs) and we have to make sure that we clean everything up eventually 🤔 Using temporary tables performed better than I expected ([source](https://github.com/AlexandraImbrisca/open-MaStR/pull/11/files?short_path=4df2b3a#diff-4df2b3a5ec75e603adfe3d9621dfdf0fb126c9c5024c3e4f392fab10e25f0923))

nesnoj:
Thx for the quick update!

> * I introduced 2 new environment variables: one for using the recommended number of processes, one to set up a custom number of processes. I think that 2 variables are necessary since people might not be aware of what number of processes would perform the best, but it would be nice to allow them to customize it

I'll get back to this later

> * @nesnoj I think the "duplicate column name" exception occurs because of a race condition (i.e., 2 processes trying to add the same column at the same time). Please correct me if I'm wrong, but I think we can safely ignore this error since once we have introduced the missing columns, we reached our purpose 🤔 I added some more error handling. Could you please let me know if you are still able to reproduce this issue?
> * I fixed the PostgreSQL issue and generally tested more for PostgreSQL

The column issue seems to be solved but now I keep getting an error in PostgreSQL with the privileges, see below for full log. The user has all privileges for the DB (superuser) and the tables are created but no data is written. I think it is not related to the actual privileges but the implementation but I wasn't able to track it further down right now.
Does it work properly at your end?

> * I updated the documentation and added a message to promote this feature
> 
> About merging the DBs: that might work, but it might get quite messy with many processes (i.e., we could end up with 10+ temporary DBs) and we have to make sure that we clean everything up eventually 🤔 Using temporary tables performed better than I expected ([source](https://github.com/AlexandraImbrisca/open-MaStR/pull/11/files?short_path=4df2b3a#diff-4df2b3a5ec75e603adfe3d9621dfdf0fb126c9c5024c3e4f392fab10e25f0923))

Great that you already did some testing in the past! The write-temp-and-merge strategy was just a quick thought, it probably comes with other consequences I cannot estimate and also requires more testing. I'm also fine with the current implementation but open for discussion ;).

<details>
<summary>Click here for full postgres traceback</summary>

```
Processing file 'AnlagenEegSolar_48.xml'...
Processing file 'EinheitenSolar_48.xml'...

concurrent.futures.process._RemoteTraceback: 
"""
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 146, in __init__
    self._dbapi_connection = engine.raw_connection()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3298, in raw_connection
    return self.pool.connect()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 449, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 1263, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 712, in checkout
    rec = pool._do_get()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/impl.py", line 179, in _do_get
    with util.safe_reraise():
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/util/langhelpers.py", line 146, in __exit__
    raise exc_value.with_traceback(exc_tb)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/impl.py", line 177, in _do_get
    return self._create_connection()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 390, in _create_connection
    return _ConnectionRecord(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 674, in __init__
    self.__connect()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 900, in __connect
    with util.safe_reraise():
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/util/langhelpers.py", line 146, in __exit__
    raise exc_value.with_traceback(exc_tb)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 896, in __connect
    self.dbapi_connection = connection = pool._invoke_creator(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/create.py", line 646, in connect
    return dialect.connect(*cargs, **cparams)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 622, in connect
    return self.loaded_dbapi.connect(*cargs, **cparams)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"
connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"


The above exception was the direct cause of the following exception:

  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/concurrent/futures/process.py", line 246, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/nesnoj/git-repos/OpenEnergyPlatform/open-MaStR/open-MaStR_546_parsing_speed/open_mastr/xml_download/utils_write_to_database.py", line 103, in process_xml_file
    create_database_table(engine, xml_table_name)
  File "/home/nesnoj/git-repos/OpenEnergyPlatform/open-MaStR/open-MaStR_546_parsing_speed/open_mastr/xml_download/utils_write_to_database.py", line 215, in create_database_table
    orm_class.__table__.drop(engine, checkfirst=True)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/sql/schema.py", line 1299, in drop
    bind._run_ddl_visitor(ddl.SchemaDropper, self, checkfirst=checkfirst)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3248, in _run_ddl_visitor
    with self.begin() as conn:
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3238, in begin
    with self.connect() as conn:
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3274, in connect
    return self._connection_cls(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 148, in __init__
    Connection._handle_dbapi_exception_noconnection(
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2439, in _handle_dbapi_exception_noconnection
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 146, in __init__
    self._dbapi_connection = engine.raw_connection()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 3298, in raw_connection
    return self.pool.connect()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 449, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 1263, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 712, in checkout
    rec = pool._do_get()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/impl.py", line 179, in _do_get
    with util.safe_reraise():
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/util/langhelpers.py", line 146, in __exit__
    raise exc_value.with_traceback(exc_tb)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/impl.py", line 177, in _do_get
    return self._create_connection()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 390, in _create_connection
    return _ConnectionRecord(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 674, in __init__
    self.__connect()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 900, in __connect
    with util.safe_reraise():
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/util/langhelpers.py", line 146, in __exit__
    raise exc_value.with_traceback(exc_tb)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/pool/base.py", line 896, in __connect
    self.dbapi_connection = connection = pool._invoke_creator(self)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/create.py", line 646, in connect
    return dialect.connect(*cargs, **cparams)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 622, in connect
    return self.loaded_dbapi.connect(*cargs, **cparams)
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/site-packages/psycopg2/__init__.py", line 122, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"
connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"

(Background on this error at: https://sqlalche.me/e/20/e3q8)
"""

The above exception was the direct cause of the following exception:

  File "/home/nesnoj/git-repos/OpenEnergyPlatform/open-MaStR/open-MaStR_546_parsing_speed/testing.py", line 17, in <module>
    db.download(data="solar")# solar
  File "/home/nesnoj/git-repos/OpenEnergyPlatform/open-MaStR/open-MaStR_546_parsing_speed/open_mastr/mastr.py", line 244, in download
    write_mastr_xml_to_database(
  File "/home/nesnoj/git-repos/OpenEnergyPlatform/open-MaStR/open-MaStR_546_parsing_speed/open_mastr/xml_download/utils_write_to_database.py", line 65, in write_mastr_xml_to_database
    future.result()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/home/nesnoj/miniconda3/envs/py310_open_mastr_546/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"
connection to server at "localhost" (127.0.0.1), port 5432 failed: FATAL:  password authentication failed for user "mastr"

(Background on this error at: https://sqlalche.me/e/20/e3q8)

Process finished with exit code 1
```
</details>

AlexandraImbrisca:
Thanks a bunch for finding this bug! I was using an unauthenticated database and I didn't realise that this could be an issue. The connection_url obfuscates the password so I updated the code to properly set the password. Could you please try again and let me know if you see the same issue?

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -36,34 +43,146 @@ def write_mastr_xml_to_database(
...
     print("Bulk download and data cleansing were successful.")
 
 
+def get_number_of_processes():
+    """Get the number of processes to use for the bulk download. By default, only one process is used, and we recommend
+    using the number of available CPUs - 1. If the user wants to use more processes, they can set the environment
+    variable."""
+    if "NUMBER_OF_PROCESSES" in os.environ:
+        number_of_processes = os.environ.get("NUMBER_OF_PROCESSES")
+        if number_of_processes >= cpu_count():
```
nesnoj, line 87: One more thing I forget to mention in my previous posts:
The env var `NUMBER_OF_PROCESSES` is a string causing the comparison to fail.

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -36,34 +43,146 @@ def write_mastr_xml_to_database(
...
+    connection_url: str,
+    password: str,
+    zipped_xml_file_path: str,
+    bulk_download_date: str,
+    bulk_cleansing: bool,
+) -> None:
+    """Process a single xml file and write it to the database."""
+    # If set, the connection url obfuscates the password. We must replace the masked password with the actual password.
+    if password:
+        connection_url = re.sub(r"://[^:]+:\*+@", f"://{password}@", connection_url)
```
nesnoj: With this regex the pw is not supplied at all.
```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "localhost" (127.0.0.1), port 5432 failed: fe_sendauth: no password supplied
```
It deletes the username and uses the pw as username.
Also, this solution does not allow colons in the username.

nesnoj (commented):
These two small things needed a fix, I patched..
Now it works fine with psql, thank you!

open_mastr/xml_download/utils_write_to_database.py
AlexandraImbrisca: Thanks for catching this!

open_mastr/xml_download/utils_write_to_database.py
AlexandraImbrisca: Oh, interesting! I was quite sure that the previous regex worked for me, but the new one works as well! About colons, I looked a bit and I think it might be a problem in general 🤔  I'd assume that internally the sqlalchemy Engine would need to split the connection link and it might be challenging if colons are used in the username

AlexandraImbrisca:
Thank you for spotting the issues and fixing them! If you any other suggestions, please let me know

open_mastr/xml_download/utils_write_to_database.py
nesnoj: At my end
```
s = "'postgresql+psycopg2://mastr:***@localhost:5432/mastr_test'"
pw = "mypass:23.+**2ü"
re.sub(r"://[^:]+:\*+@", f"://{pw}@", s)
```
returns `"'postgresql+psycopg2://mypass:23.+**2ü@localhost:5432/mastr_test'"` as both user+pw is replaced by pw when using `f"://{pw}@"`.
The recognition of colons in the new regex are just a nice side effect (yes, sqla might doesn't like them..)

FlorianK13:
Is this the version now that should be merged to develop and released afterwards? If yes, I would start with the comparison of the two databases:
- downloaded with this branch
- downloaded with `open-mastr` from pypi

AlexandraImbrisca:
Yes, I think this is the final version (unless we find any other bugs/suggestions). If you can help testing, that would be great! I will also test a bit more

FlorianK13:
Did you test on windows? Without setting os.environ, my program immediatly crashes:

concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.

AlexandraImbrisca:
Could you please try again and let me know if any other error is being printed? I unfortunately don't have my own Windows system and I only tested the previous version before adding the os.environ variables. I'll try accessing Windows today and test the code again

AlexandraImbrisca:
I just tested on Windows 11 and I had no issues. I tried with WSL 2.0 and similarly, the program is running correctly. I tested without setting os.environ as well as with setting each of the fields.

FlorianK13:
Just saw it now, I'll work on this hopefully within this or next week.

FlorianK13:
@AlexandraImbrisca 
Running this script:
```python
from open_mastr import Mastr
import os

os.environ["NUMBER_OF_PROCESSES"] = "1"
db = Mastr()
db.download(date="existing", data="solar")
```

## throws this error:
        concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.

        RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

        To fix this issue, refer to the "Safe importing of main module"
        section in https://docs.python.org/3/library/multiprocessing.html

## It refers to:
\xml_download\utils_write_to_database.py", line 67, in write_mastr_xml_to_database future.result()
\utils_write_to_database.py", line 63, in write_mastr_xml_to_database
    futures = [
\utils_write_to_database.py", line 64, in <listcomp>
    executor.submit(process_xml_file, *item) for item in interleaved_files

I'm using python 3.11.8 in a conda env on Windows.

AlexandraImbrisca:
Oh, could you please try again with the following snippet?

```
from open_mastr import Mastr
import os

os.environ["NUMBER_OF_PROCESSES"] = "1"
db = Mastr()

if __name__ == "__main__":
      db.download(date="existing", data="solar")
```

This condition is necessary to ensure that the program doesn't attempt to recreate the pool for every new process. It's already part of [main.py](https://github.com/OpenEnergyPlatform/open-MaStR/blob/11fb568879ae8015af74f3d6b386faa5027c2721/main.py#L73). Without this condition, multiprocessing will always break on Windows/MacOS AFAIU

FlorianK13:
This solved the issue. However I'm sure that many users don't have a ``if __name__ == "__main__":` check in their code. Is there a way to only use multiprocessing if os.environ["NUMBER_OF_PROCESSES"] = "SomeNumber" is given?

Because otherwise our version update would break their code.

nesnoj:
In my tests above it always worked without checking the top-level code environment by `__name__ == '__main__'` (linux) :thinking:

FlorianK13:
Yes, I guess this is a problem appearing only on Windows.

AlexandraImbrisca:
Great idea! I updated the code to not use multiprocessing unless one of the 2 options is set (USE_RECOMMENDED_NUMBER_OF_PROCESSES / NUMBER_OF_PROCESSES). I also added a note to the documentation about if \_\_name\_\_ == "\_\_main\_\_"

AlexandraImbrisca:
@FlorianK13 @nesnoj small reminder if you can review these changes again! 🙏🏻

nesnoj:
@FlorianK13

FlorianK13:
A simple 
```python
db = Mastr()
db.download(date="existing", data="wind")
```
on windows now works again 👍

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -36,15 +43,85 @@ def write_mastr_xml_to_database(
...
+
+def get_number_of_processes():
+    """Get the number of processes to use for the bulk download. Returns -1 if the user has not opted for the
+    parallelized implementation. Otherwise, we recommend using the number of available CPUs - 1. If the user wants to
+    use more processes, they can set the custom environment variable."""
+    if "NUMBER_OF_PROCESSES" in os.environ:
+        try:
+            number_of_processes = int(os.environ.get("NUMBER_OF_PROCESSES"))
+        except ValueError:
+            print(f"Warning: Invalid value for NUMBER_OF_PROCESSES. Fallback to 1.")
```
FlorianK13: Ruff linter gives a warning here since there is an f-string without any variable.

open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -53,15 +130,78 @@ def write_mastr_xml_to_database(
...
+        max_overflow=20,
+        # Recycle inactive connections after 180 seconds to prevent stale connections.
+        pool_recycle=180,
+        # Wait for 30 seconds before raising an exception when the pool is full.
+        pool_timeout=30,
+    )
 
-    print("Bulk download and data cleansing were successful.")
+
+def interleave_files(threads_data: []):
```
FlorianK13: ruff linter does not like this type hint -> def interleave_files(threads_data: list) should be correct.

FlorianK13 (changes requested):
Only two small changes. After that I suggest we can merge this branch and start with #602 on the development branch. Do you agree @nesnoj ?

nesnoj:
> Only two small changes. After that I suggest we can merge this branch and start with #602 on the development branch. Do you agree @nesnoj ?

yes, go for it!

AlexandraImbrisca:
Sounds great @FlorianK13, thank you! I fixed the ruff linter warnings

FlorianK13 (approved):

FlorianK13 (approved):

Issue #607, opened 2025-03-21
Upcoming MaStR release V25.1.152 - review patchnotes


nesnoj:
Neue Webdienst-Version 25.1.152
Checkout [WSDL Patchnotes V25.1.152](https://www.marktstammdatenregister.de/MaStRHilfe/files/webdienst/WSDL_Patchnotes_V25.1.152.xlsx) and [Release Notes](https://www.marktstammdatenregister.de/MaStRHilfe/subpages/releasenotes.html) for potential changes.

pt-kkraemer:
According to this

![Image](https://github.com/user-attachments/assets/6ce1b07b-66fc-4ac4-91e2-145cb812ea91)

I think the `columns_replace_list` in [columns_to_replace.py](https://github.com/OpenEnergyPlatform/open-MaStR/blob/develop/open_mastr/xml_download/colums_to_replace.py) needs to be updated.

FlorianK13:
Already got the first error in pytest: ERROR tests/test_mastr.py::test_Mastr_translate - TypeError: 'TechnologieFlugwind' is an invalid keyword argument for WindExtended

FlorianK13:
@pt-kkraemer If you are interested in contributing, you can change the columns in [columns_to_replace.py](https://github.com/OpenEnergyPlatform/open-MaStR/issues/open_mastr/xml_download/colums_to_replace.py) and create a PR. We are always happy about contributions from outside.

pt-kkraemer:
Yes, I'll do that

FlorianK13:
I tested the download with the changes from #611 now and it works. Hence the changes to the data model of MaStR from this issue do not need to be merged as a hotfix on production but we can merge them on develop.

pt-kkraemer:
I just downloaded the new bulk data and was first very happy because it was a very fast download but after opening some of the xml-Files I fund the following (and nothing else) inside:

`EinheitenWind.xml`:

> Msg 207, Level 16, State 1, Server VS1301, Line 1
> Invalid column name 'WindAnLandOderAufSee'.
> Msg 207, Level 16, State 1, Server VS1301, Line 1
> Invalid column name 'TechnologieFlugwindenergieanlage'.
> Msg 207, Level 16, State 1, Server VS1301, Line 1
> Invalid column name 'Flughoehe'.
> Msg 207, Level 16, State 1, Server VS1301, Line 1
> Invalid column name 'Flugradius'.

[...]

PR #613, opened 2025-04-03
611 multiple mastr download links

### LINKED ISSUE #611: Wrong url for bulk download close to MaStR release dates
In the function `gen_version` of `utils_download_bulk.py` the version for the bulk download is not correctly created.

The current download link is 

> https://download.marktstammdatenregister.de/Gesamtdatenexport_20250403_24.2.zip

whereas the function creates

> 25.1

I guess this problem will resolve tomorrow according to their [Release-Termine](https://www.marktstammdatenregister.de/MaStRHilfe/files/webdienst/Release-Termine.pdf).

Maybe if an error is raised that the link cannot be found one could automatically try with another version?

FlorianK13:
## Summary of the discussion

Around the time of new releases, the MaStR version number in the download link might be wrong. This PR adds some robustness by also checking download links with the MaStR version that was used before the current one and version that will be used next.

Old: Only check Download link for 23.1
New. First check 23.1, then check 22.2, then check 23.2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # own testing files
 tmptest.py
 
+# uv and rye
+uv.lock
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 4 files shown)


nesnoj (approved):
Nice, thx @FlorianK13 !
I just adjusted some small things.
I guess the failing test related to "TechnologieFlugwind" is subject of #607

PR #621, opened 2025-04-15
Add new columns to orm.py #620

### LINKED ISSUE #620: API download fails since BNetzA updates
When running
```python
db_api = Mastr()
db_api.download(method="API", api_limit=10)
```

I get the error:
TypeError: 'TechnologieFlugwind' is an invalid keyword argument for WindExtended

I think the solution lies in updating `orm.py` and I will investigate this further.

FlorianK13:
- Added new lines to orm.py
- Changed the `translate` test to use a database created from xml files, not the API database. This was needed because otherwise the tests were failing.


nesnoj (commented):
Executing
```
from open_mastr import Mastr
db = Mastr()
db.download(method="API", api_limit=2)
```
I get
```
[...]
Downloading permit_unit_data (wind): 100%|██████████| 2/2 [00:01<00:00,  1.65unit/s]
[...]
TypeError: 'DatumAntragstellung' is an invalid keyword argument for Permit
```
I cannot find this column anywhere in the repo. Did they change sth overnight?

FlorianK13:
From the downloaded xml files following new attribute was introduced: permit.DatumDerAntragstellung

In the bulk files, this column is named DatumDerAntragstellung. Since method=API is deprecated I'd say we keep the bulk download column name?

nesnoj (approved):
Still fails due to this error #620 but as API download is deprecated #622 we won't fix this.

PR #666, opened 2025-10-20
Improve logging

### LINKED ISSUE #657: consistent usage of logging instead of print
## Description of the issue
I would like to have a logfile created with all the output from the download call to open-mastr.
Instead, this package is currently heavily using print instead.

## Ideas of solution

Switch to logging where possible and add an option to silence the `tqdm`.

## Workflow checklist

### LINKED ISSUE #664: Importing open-mastr changes logging settings
## Description of the issue

In https://github.com/OpenEnergyPlatform/open-MaStR/blob/develop/open_mastr/__init__.py things are happening from a plain import - which should not be the case.
Especially when using this library as part of something else.

```
import logging
logging.debug("hello")
```
should not print anything, as the default logger is set to warning.
With a `import open_mastr` before, we do see all debug logs of urllib and everything else, which should not be the case.

## Ideas of solution

only adjust logging settings if the cli tool is used

## Context and Environment
* Version used: 
* Operating system: 
* Environment setup and (python) version: 

## Workflow checklist
- [ ] I have checked [the documentation](https://open-mastr.readthedocs.io/en/latest/) and confirmed this issue is not already addressed there.
- [ ] I am aware of the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)

nesnoj:
## Summary of the discussion

- Replace print() statements by logging #657
- Do not set root logger when importing open-mastr #664
- Set default log level to INFO
- Extends logging

### Automation
Fixes #657 
Fixes #664

### PR-Assignee

### Reviewer

### DIFF
--- a/open_mastr/soap_api/metadata/description.py
+++ b/open_mastr/soap_api/metadata/description.py
@@ -1,9 +1,12 @@
 from io import BytesIO
+import logging
 import re
 from urllib.request import urlopen
 from zipfile import ZipFile
 import xmltodict
 from collections import OrderedDict
+
+log = logging.getLogger(__name__)
 
 
 class DataDescription(object):
@@ -150,9 +153,11 @@ class DataDescription(object):
                             fcn["sequence"]["element"]["@type"].split(":")[1]
                         ]["sequence"]["element"]
                 else:
-                    print(type(fcn["sequence"]))
-                    print(fcn["sequence"])
-                    raise ValueError
+                    log.error(f"Unexpected sequence type: {type(fcn['sequence'])}")
+                    log.error(f"Sequence content: {fcn['sequence']}")
+                    raise ValueError(
+                        f"Unexpected sequence structure in function metadata"
+                    )
 
                 # Add data for inherited columns from base types
                 if "@base" in fcn:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
   [#636](https://github.com/OpenEnergyPlatform/open-MaStR/pull/636)
 - Change print statement about data cleansing
   [#650](https://github.com/OpenEnergyPlatform/open-MaStR/pull/650)
+- Improve logging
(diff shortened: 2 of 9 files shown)


nesnoj:
@maurerle could you check if this works for you?

maurerle (approved):
I looked through the code. I think this is a better and more complete approach.
Thanks!!

open_mastr/soap_api/metadata/description.py
```diff
@@ -1,10 +1,13 @@
 from io import BytesIO
+import logging
 import re
 from urllib.request import urlopen
 from zipfile import ZipFile
 import xmltodict
 from collections import OrderedDict
 
+log = logging.getLogger(__name__)
```
FlorianK13, line 9: @nesnoj can you shortly state the difference between this method and the log=setup_logger() method as it is used in [utils_write_to_database.py](https://github.com/OpenEnergyPlatform/open-MaStR/pull/666/files#diff-5a2d70504731a2e36460e60408ece91491dd1aacb3011f504f333e5e08ec4aff)

FlorianK13 (dismissed):

open_mastr/soap_api/metadata/description.py
maurerle: logging.getLogger does return the logger with the name of the current file.
In this case this would be `open_mastr.soap_api.metadata.description`.

This btw allows for a very fine-grained setting of the log settings as well (for example disabling only the soap part using `logging.getLogger("open_mastr.soap_api").setLevel("ERROR")`).

On the other hand, the `setup_logger` function also calls `logging.config.dictConfig(logging_config)` which configures the log settings of the user according to the dict (which is not intended when just importing the module)

maurerle:
@nesnoj I just tested it as well - works well :)

open_mastr/soap_api/metadata/description.py
nesnoj: Yes, thx for the explanation @maurerle

PR #667, opened 2025-10-21
Improvements in handling bulk downloads

### LINKED ISSUE #564: Prior bulk zip files are deleted on download
Alle files in the xml dir [are cleared](https://github.com/OpenEnergyPlatform/open-MaStR/blob/e03e824885d4f421c30aeb7c97f59f2570bea9ef/open_mastr/xml_download/utils_download_bulk.py#L112-L113) when a download is triggered. This is annoying when users want to keep old releases.
The xml dir needs to be created though, but this is redundantly done [before the download starts](https://github.com/OpenEnergyPlatform/open-MaStR/blob/e03e824885d4f421c30aeb7c97f59f2570bea9ef/open_mastr/mastr.py#L228).

Is there a reason why it is done this way - any objections against removing this behaviour? @FlorianK13

### LINKED ISSUE #668: Zip file is deleted on full bulk download
Zip file from download is deleted every time a full bulk download is started, no matter if it's from today or outdated. However, in partial bulk download an existing file is detected and used.
This bug was introduced in #652

I'll include a fix in #667

nesnoj:
## Summary of the discussion

- Introduce parameter to allow keeping old XML files #564 
- Prevent deletion of up-to-date zip file on full bulk download #668
- Add technology checks to full bulk download -> do not download if data is present; fixed behavior for full and partial download:
  - zip file **does** contain all technologies -> use existing zip file
  - zip file **does not** contain all requested technologies -> download full / partial

### Automation
Closes #564
Fixes #668

### PR-Assignee

### Reviewer

### DIFF
--- a/open_mastr/xml_download/utils_download_bulk.py
+++ b/open_mastr/xml_download/utils_download_bulk.py
@@ -123,7 +123,11 @@ def download_xml_Mastr(
     Parameters
     -----------
     save_path: str
-        The path where the downloaded MaStR zipped folder will be saved.
+        Full file path where the downloaded MaStR zip file will be saved.
+    bulk_date_string: str
+        Date for which the file should be downloaded.
+    xml_folder_path: str
+        Path where the downloaded MaStR zip file will be saved.
     """
 
     print_message = "Starting the Download from marktstammdatenregister.de."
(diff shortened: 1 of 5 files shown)


open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -197,28 +203,59 @@ def check_download_completeness(
     return list(missing_data_set), is_katalogwerte_existing
 
 
-def delete_xml_files_not_from_given_date(save_path: str, xml_folder_path: str):
+def delete_xml_files_not_from_given_date(
```
FlorianK13, line 207: The naming here is very unintuitive with the new functionality. The method says it deletes the xml files, but then it has an attribute where you can skip the deletion. Maybe the 
```python
if not keep_old_downloads:
```
should be moved to the parent method? So that `delete_xml_files_not_from_given_date` is only called if the xml files should be deleted?

FlorianK13 (commented):
Left a first comment here, but I haven't finished the review yet (was running out of time)

open_mastr/xml_download/utils_download_bulk.py
nesnoj: Yes, rookie mistake! I moved it outside of the function..

FlorianK13 (approved):
Looks good to me!

PR #671, opened 2025-10-21
Modernize linting setup and add pre-commit checks in CI

### LINKED ISSUE #584: Extend linters in precommit hook
Currently in the precommit hooks only black is applied which is insufficient.

nesnoj:
_**AI disclosure:** The code, commit and PR texts have been created with support of AI._

## Summary of the discussion

The linting setup of open-MaStR was outdated: `flake8`, `black` and `pylint` were configured separately, `pre-commit` only ran `black`, and nothing checked linting in CI.

This PR modernizes the linting and code quality setup for the project using ruff. It also establishes automated pre-commit checks in CI (new workflow: `.github/workflows/pre-commit.yml`) to ensure consistent code quality across the workflow -> see passing action below.

- You can check it by reinstall open-mastr (`pip install -e .[dev]`) and committing something or run `pre-commit run --all-files`
- On commit, the pre-commit hooks should be automatically installed 
- Currently some directories are excluded from the tests
  - `postprocessing/` and `open_mastr/soap_api/`: will be removed soon so there's no need to bother about the code quality right now
  - `open_mastr/resources`: holds the fallback XSDs
  - `scripts/utils/data_io.py`: unused, will be removed in the future
- The detailed config can be found below
- The **linting has been applied to the entire repo** in the following commits, the introduced CI action proves all tests passing. However, the application to the codebase required changes which are more than cosmetic, details see below.
  - 06df67ae788348d8a20035d77d980e2c27aa12b1
  - 3b5c9564a568587720a6284d88fc5fdfc9349039
  - ec7be3cd26e86d650657ab03f401b857dc1f5f31

Tested:
- All tests pass
- Full bulk download and CSV export run successfully
- Script `release_test.py` passes, except for the issue with "Ertuechtigungsart" described in https://github.com/OpenEnergyPlatform/open-MaStR/pull/801#issuecomment-5700044039

### Configuration

- `ruff` with `E`, `F`, `I`, `N`, `W`, `UP` and `D` (numpy docstring convention). `D1` (missing docstrings) is ignored — the rules require existing docstrings to be well-formed, but do not force a docstring onto every function.
- The formatter wraps at 88 characters, while `E501` only complains above 100 (the limit the project used with flake8). Without that split, every long string, URL and comment that the formatter cannot break would be reported.
- `open_mastr/soap_api/`, `open_mastr/resources/`, `postprocessing/` and the single file `scripts/utils/data_io.py` are excluded; everything else, `scripts/` included, is checked. The exclusions live in `pyproject.toml` and are mirrored in `.pre-commit-config.yaml`; `force-exclude` makes them apply to the file paths that pre-commit passes explicitly, so a local `ruff check .` sees exactly the same files as the hook.
- `open_mastr/resources/` is excluded from all hooks because it holds the vendored BNetzA fallback XSD files, which `scripts/update_fallback_xsd.py` regenerates verbatim. Without the exclusion, every refresh would produce a whitespace diff.
- `check-yaml` runs with `--unsafe`, because `mkdocs.yml` uses the MkDocs Material tag `!!python/name:material.extensions.emoji.twemoji`, which safe-load rejects.

### Application to the codebase

229 ruff findings and 31 mypy errors were addressed. Most were mechanical, but a few were real defects:

- `id_column_name` in `sqlalchemy_tables.py` was dead since a rename.
- `get_mastr_token` swallowed every exception with a bare `except`.
- `shutil.Error` was imported only to annotate two functions that actually receive `sqlalchemy.exc.DataError` and `lxml.etree.XMLSyntaxError`.
- `download_documentation` called `requests.get(None)` when the preferred URL returned 404 and no fallback URL existed — which is the case whenever a custom URL is passed. The user got a `TypeError` instead of the actual HTTP error.
- `transform_date_parameter` advertised support for `datetime.date`, which `parse_date_string` cannot process.
- `create_bulk_download_mockup.py` passed `root_tag` to `etree.Element()` without checking it, so a table without any XML record raised a `TypeError` instead of naming the table.
- `scripts/utils/data_io.py` cannot be imported at all: it needs the undeclared dependency `pynodo` and two names that no longer exist (`create_datapackage_meta_json`, removed from `soap_api/metadata/create.py`, and `get_data_version_dir`, removed in #748). Nothing in the repository imports it and it has not been touched since 2022, so it is excluded from the hooks rather than repaired.

All 64 tests pass unchanged. The `%`-formatting converted to f-strings in `format_sqlalchemy_column` / `format_sqlalchemy_table` was verified to produce identical output, and long string literals were only split with implicit concatenation, so no string content changed.

### Known limitations

- The `pre-commit` workflow is triggered on `push` and `pull_request` for `production` and `develop`.
- One mypy error is suppressed rather than fixed: `translate_mastr_column_name` returns `None` for untranslatable primary key columns, which then match no column and leave the table with the artificial primary key. Documented with a TODO. Two further errors disappear only because `scripts/utils/data_io.py` is excluded.
- `open_mastr/soap_api/` stays unlinted. It contains 42 ruff findings, among them a call to `create_datapackage_meta_json` in `soap_api/metadata/create.py`, which is commented out at the definition site and would raise at runtime.
- `scripts/utils/data_io.py` is excluded rather than fixed or deleted, so the three breakages described above stay in the tree.
- The docstring of `Mastr.download` still documents `datetime.datetime` for `date`, which does not work.

### Added
- Modernize linting setup and add pre-commit checks in CI - Remove the flake8, black and pylint configuration and development dependencies and add ruff and mypy.
  [(#671)](https://github.com/OpenEnergyPlatform/open-MaStR/pull/671)

### Changed
- Apply the new ruff and mypy rules to the codebase
  [(#671)](https://github.com/OpenEnergyPlatform/open-MaStR/pull/671)

### Automation
Closes #584

### PR-Assignee

### Reviewer

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,32 @@
+name: Pre-commit Checks
+
+on:
+  pull_request:
+    branches:
+      - production
+      - develop
+  push:
+    branches:
+      - production
+      - develop
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    name: Run pre-commit hooks
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pre-commit
+
+      - name: Run pre-commit
+        run: pre-commit run --all-files --show-diff-on-failure
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.15.0
+current_version = 0.16.0
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)((?P<release>(a|na))+(?P<build>\d+))?
 serialize = 
 	{major}.{minor}.{patch}{release}{build}
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,0 @@
-[flake8]
-exclude = meta/migrations/
-max-line-length = 100
-extend-ignore = E203
--- a/.github/workflows/ci-develop.yml
+++ b/.github/workflows/ci-develop.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: ${{ !github.event.pull_request.draft }}
     strategy:
+      max-parallel: 1
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12']
--- a/.github/workflows/ci-production.yml
+++ b/.github/workflows/ci-production.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: ${{ !github.event.pull_request.draft }}
     strategy:
+      max-parallel: 1
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12']
--- a/.github/workflows/ci-production.yml
+++ b/.github/workflows/ci-production.yml
@@ -33,7 +33,7 @@ jobs:
     - name: create package
       run: python -m build --sdist
     - name: import open-mastr
-      run: python -m pip install ./dist/open_mastr-0.15.0.tar.gz
+      run: python -m pip install ./dist/open_mastr-0.16.0.tar.gz
     - name: Create credentials file
       env:
         MASTR_TOKEN: ${{ secrets.MASTR_TOKEN }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
 
   # Ruff - fast, comprehensive linter and formatter (replaces flake8, isort, pylint, pyupgrade, black, pydocstyle)
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.1
+    rev: v0.14.9
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
@@ -32,7 +32,7 @@ repos:
 
   # mypy - static type checking
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.18.2
+    rev: v1.19.1
     hooks:
       - id: mypy
         additional_dependencies:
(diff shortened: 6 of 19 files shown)


FlorianK13:
I remember that @nesnoj expects a larger diff when linting the whole repo. We used black until now - will there be such a large difference in the codebase between ruff and black linting?

nesnoj:
> I remember that @nesnoj expects a larger diff when linting the whole repo. We used black until now - will there be such a large difference in the codebase between ruff and black linting?

Yes, definitely! Black only covers some aspects of code quality. Furthermore, in this case ruff is configured quite extensively.
I showcased the changes in #694 .
The most important question is: when do we want to lint the codebase? I propose to do it as last step before the next release.

nesnoj:
I updated the branch with `develop`, thx for taking over the review @Simon-Will :)
It's definitely one of the most extensive in this repo..

nesnoj:
I tested again with today's data:
DB download, CSV export, tests and `release_test.py` works out.

Tests returns warnings as described in #781.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+    branches:
+      - production
+      - develop
+  push:
+    branches:
+      - production
+      - develop
```
FlorianK13:
```suggestion
```

I think it is enough if the checks run on the PR, they do not need to run again if the PR is merged.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    name: Run pre-commit hooks
+
+    steps:
+      - uses: actions/checkout@v4
```
FlorianK13:
```suggestion
      - uses: actions/checkout@v7
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
```
FlorianK13:
```suggestion
        uses: actions/setup-python@v7
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          python-version: '3.11'
```
FlorianK13:
```suggestion
          python-version: '3.14'
```

FlorianK13: Or is v3.11 needed for pre-commit here?

FlorianK13 (changes requested):

.github/workflows/pre-commit.yml
nesnoj: No, 3.14 is fine.

nesnoj:
I applied your changes, thx for checking @FlorianK13 !

FlorianK13 (approved):
LGTM

Simon-Will:
Thanks, @nesnoj for implementing and @FlorianK13 for reviewing! I think this is really an important step forward to make development easier for us. 🥳

Issue #663, opened 2025-09-29
Automatically read the data download export description to extract all value mappings


pt-kkraemer:
## Description of the issue

When downloading the bulk data some non-integer values are stored as integers in order to reduce file size. Most of these integer-text-mappings can be found in the `Katalogwerte.xml` from the bulk download. But some of the mapping values are not present there, see for example #653.

## Ideas of solution

We could try to download the export description PDF from the bulk download which can be found [here](https://www.marktstammdatenregister.de/MaStR/Datendownload) and use that.

## Workflow checklist
- [ ] Find out if the link of the export description has a pattern or if it changes 
- [ ] Find a suitable python package to read PDF or extract tables from PDF
- [ ] Update the values in [colums_to_replace.py](https://github.com/OpenEnergyPlatform/open-MaStR/blob/develop/open_mastr/xml_download/colums_to_replace.py)
- [ ] I am aware of the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)

FlorianK13:
Do you intend this as part of the python package? Or as a script to create PRs for updates of the python package?

nesnoj:
- Min solution: Alert message only
- Better: Bot via GH action that compares values from repo with (downloaded and scraped) the PDF and creates a PR

nesnoj:
> * Better: Bot via GH action that compares values from repo with (downloaded and scraped) the PDF and creates a PR

Unfortunately I wasn't able to find an alternative source for the Katalogwerte (they are not contained in the PDF).

I see 2 checks that could be done:
1. Verification of columns to be replaced: compare XSD files with [`columns_replace_list`](https://github.com/OpenEnergyPlatform/open-MaStR/blob/c8c16ce68dc4406dd8f4d4df88a22362ba4de1f5/open_mastr/xml_download/colums_to_replace.py#L46)
Example: "Hauptbrennstoff" in `EinheitenVerbrennung.xsd` -> yes, included in `columns_replace_list`
2. Verification if columns from 1. to be mapped are included in `Katalogwerte.xml`
Example: "Hauptbrennstoff" in `EinheitenVerbrennung.xsd` -> yes, included in `Katalogwerte.xml`

If these 3 sources are not in sync, a PR (or at least an issue) is created.

**On the value mapping:**
[...]

Issue #516, opened 2024-04-25
Dynamically add new tables to existing databases


FlorianK13:
If new tables are introduced by BNetzA, the current code cannot simply add them to an existing database. For example, when running `db.download()` and there is already a database initialized, a new table will cause an error. This should be changed, ie. new tables should be added automatically.

They should take the name from `orm.py` if the new table is already part of a new open-mastr release. Or they should just use the german name otherwise.

nesnoj:
An automatic creation be great, but the value replacement (e.g. 4=biomass) should be added too.

FlorianK13:
If only this was already implemented, we would not need to handle #572 under time pressure 🙄

FlorianK13:
With 0.14.5 we now have at least the case that tables that are new do not result in a crash anymore. However they are still not automatically added.

nesnoj:
Yes, let's talk about this in our JF

nesnoj:
Notes from JF:
- Dynamic update needed
- There seems to be no way of getting the structure in machine-readable form prior to the public release. So no early adapted release is possible.
- How to proceed with orm.py?
  - Regularly parse xsd from ["Beschreibung des Exports"](https://www.marktstammdatenregister.de/MaStR/Datendownload)  and create PR with ORM update?
  - Get rid of orm.py - only dynamic creation
  - keep in mind: existing DB structure and CSV export

nesnoj:
As soon as #635 is solved, we could switch to dynamic creation.
Open question: Are the provided xsd files always in sync with the Gesamtdatenauszug? We could store the xsd set as fallback in the repo though.

Simon-Will:
Hi, I'm interested in contributing to this project. (Also attended the MaStR4All workshop this week.) I opened #675 and accompanying pull request #676 recently, which would prevent tables from being created automatically. In that pull request, @FlorianK13 briefly mentioned creating tables dynamically, which is probably this issue.

### XSD files

> Open question: Are the provided xsd files always in sync with the Gesamtdatenauszug?

Hard to answer in a general, future-proof way, but I inspected the current (2025-11-29) MaStR export & XSD files. Using `xmllint`, I validated each file against its schema and found countless violations. That's annoying, but not a blocker for this issue here since most fall into one of these categories:
* XSD doesn't mark attributes with `minOccurs="0"` even if some XML entities lack the attributes.
* XSD demands an attribute order slightly different from the one in the XMLs.
[...]

PR #676, opened 2025-11-06
Add option to not create database tables

### LINKED ISSUE #675: Give the option of expecting tables to be present instead of creating them
## Description of the issue

We're trying to load the MaStR data for solar plants only into an existing MySQL database. Since we want to retain control of our database state and have changes be done via [alembic](https://alembic.sqlalchemy.org/en/latest/) migrations, we would like to create the tables before instantiating the `Mastr` object.

We created the tables necessary for the solar download (`solar_extended`, `solar_eeg`) ourselves, but of course `Mastr` tries to create the others. Since we (unfortunately) use MySQL, which unfortunately requires an explicit length for varchar fields (i.e. `String(length=some_len)` instead of bare `String`), instantiating `Mastr` fails with:

```
sqlalchemy.exc.CompileError: (in table 'basic_units', column 'EinheitMastrNummer'): VARCHAR requires a length on dialect mysql
```

## Ideas of solution

Our issue has two aspects:
* MySQL is not supported.
* One cannot prevent `Mastr` from trying to create tables.

One easy fix for us would be **making it possible to skip the table creation step**. This is in general a good idea in my opinion to give the user more control over their database if desired. In some environments, this is also necessary because runtime database users aren’t granted permission to use any DDL statements for [security reasons](https://cheatsheetseries.owasp.org/cheatsheets/Database_Security_Cheat_Sheet.html#creating-secure-permissions). (One can exploit things like function creation in Postgres to get access to the filesystem, arbitrary commands, etc.)

Another step would be supporting MySQL. This requires filling out all the length attributes of the `String` fields, which is definitely annoying work and may not be very easy to do in a future-proof way.

A third step would be instructing `Mastr` to only create those tables that are necessary for the intended download.

## Workflow checklist

Simon-Will:
## Summary of the discussion

I propose to introduce this little option for `Mastr` that enables the user to prevent it from creating tables automatically. If you agree this is a sensible addition, I can also add some documentation for it.

### Added
- Add new parameter `ensure_tables` with default value `True` to class `Mastr` that controls whether to create database tables

### Automation
Closes #675

### PR-Assignee

### Reviewer

### DIFF
--- a/open_mastr/mastr.py
+++ b/open_mastr/mastr.py
@@ -69,14 +69,20 @@ class Mastr:
         Defines the engine of the database where the MaStR is mirrored to.
         Default is 'sqlite'.
     connect_to_translated_db: boolean, optional
-            Allows connection to an existing translated database. Default is 'False'.
-            Only for 'sqlite'-type engines.
-
-
-
+        Allows connection to an existing translated database. Default is 'False'.
+        Only for 'sqlite'-type engines.
+    create_and_alter_database_tables: boolean, optional
+        Automatically creates the database tables necessary for storing the MaStR data.
+        Default is 'True'. Set this to 'False' if you prepare the database for the download
+        yourself and don't want this class to touch your database definitions.
     """
 
-    def __init__(self, engine="sqlite", connect_to_translated_db=False) -> None:
+    def __init__(
+        self,
+        engine="sqlite",
+        connect_to_translated_db=False,
+        create_and_alter_database_tables=True,
+    ) -> None:
         validate_parameter_format_for_mastr_init(engine)
 
         self.output_dir = get_output_dir()
@@ -102,7 +108,9 @@ class Mastr:
             "'pip install --upgrade open-mastr'\n"
         )
 
-        orm.Base.metadata.create_all(self.engine)
+        self.create_and_alter_database_tables = create_and_alter_database_tables
+        if self.create_and_alter_database_tables:
+            orm.Base.metadata.create_all(self.engine)
 
     def download(
         self,
@@ -254,6 +262,7 @@ class Mastr:
                 data=data,
                 bulk_cleansing=bulk_cleansing,
                 bulk_download_date=bulk_download_date,
+                create_and_alter_database_tables=self.create_and_alter_database_tables,
             )
 
         if method == "API":
@@ -417,7 +426,15 @@ class Mastr:
             print(df.head(10))
             ```
 
+        This method will only work with SQLite databases and if :class:`Mastr`
+        is constructed with :attr:`Mastr.create_and_alter_database_tables` set
+        to False.
         """
+        if not self.create_and_alter_database_tables:
+            raise ValueError(
+                "Translating the database always includes altering tables."
+                " So this is incompatible with the option `create_and_alter_database_tables`."
+            )
 
         if "sqlite" not in self.engine.dialect.name:
             raise ValueError("engine has to be of type 'sqlite'")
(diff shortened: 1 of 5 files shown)


open_mastr/mastr.py
```diff
@@ -76,7 +76,9 @@ class Mastr:
 
     """
 
-    def __init__(self, engine="sqlite", connect_to_translated_db=False) -> None:
+    def __init__(
+        self, engine="sqlite", connect_to_translated_db=False, ensure_tables=True
```
FlorianK13: `ensure_tables` is not really self-explanatory. Maybe `create_empty_tables`?

open_mastr/mastr.py
```diff
@@ -76,7 +76,9 @@ class Mastr:
```
FlorianK13: A documentation of this new parameter is missing. @Simon-Will Can you add it here? It will than automatically be displayed in most code editors and on our documentation page.

FlorianK13 (commented):
Hi @Simon-Will, thanks for your contribution. I have added two comments, maybe you can adapt it that way?

open_mastr/mastr.py
Simon-Will: Yes, I can do that later.

Simon-Will:
Actually, it turned out that what I did here is not enough: There are other places where existing tables are dropped and a new creation is attempted: At least here: https://github.com/OpenEnergyPlatform/open-MaStR/blob/d2bcd43cc0a25ed03e687224879f411e3119b065/open_mastr/xml_download/utils_write_to_database.py#L240

It's probably for ensuring that old data is deleted, right?

I will see if I find a reasonable way to adapt this.

Simon-Will:
I updated this PR and think that it's ready now. Could you have another look, @FlorianK13? I was a bit surprised to find some ALTER statements quite deep in the code. This meant that I also had to pass along the new switch quite deeply.

I think it's alright for now, but do you think it makes sense to concentrate the table creation in one point: After downloading the Gesamtdatenexport, one could quickly scan for columns not present in the SQLAlchemy models. And only then create the tables. When the tables are created, one could start importing.

FlorianK13:
@Simon-Will The amount of technical debt in this repository is always good for a surprise or two. Regarding the creation at one place - we are somehow touching this topic in our discussion in #663
There would also be the possibility to create tables based on the `.xsd` files from the `Beschreibung des Exports` from [here](https://www.marktstammdatenregister.de/MaStR/Datendownload) - instead of using the `orm.py`

Simon-Will:
@FlorianK13, is there anything else you'd like me to do here? Or is it too deep a change in general?

FlorianK13:
@Simon-Will I haven't found the time yet to review your PR since I have only limited resources for open-mastr. I'll ping you if I think more changes are necessary.

Simon-Will:
@FlorianK13, yes, perfectly fine! I didn't want to put pressure on you; was just trying to make sure that you're not currently waiting for anything from my side. Take your time. :)

FlorianK13:
Hi @Simon-Will - as a short update: @nesnoj plans to make a release next week, but we have not really worked on your PR yet. So your PR will very likely not be part of the release next week. Since we have more PRs waiting to be merged that will not make it into the release next week, we hopefully manage to have another release soon.

FlorianK13:
I really hope this is not too demotivating for you :/ It is just that me and the other maintainers really struggle with their time budget and (sadly) open-mastr is not always first priority.

Simon-Will:
@FlorianK13, thanks for the heads-up! That's understandable and quite alright. See you at the workshop at EWI next week in case you're there!

FlorianK13:
No, I won't make it - but I wish you a nice time at the workshop (:

FlorianK13:
Converted to draft, since the work in #516 could make this PR to be irrelevant

FlorianK13:
@Simon-Will I think this is not necessary anymore after #718 - or am I wrong?

Simon-Will:
Indeed. I'm closing it.

PR #684, opened 2025-12-07
516: Dynamically create database tables based on XSD files

### LINKED ISSUE #516: Dynamically add new tables to existing databases

Simon-Will:
## Summary of the discussion

A first attempt at creating tables based on the MaStR XSD files.

Checklist until this is really ready:
- [x] Download current documentation and create database model from it
- [x] Use new database model for all the insertion code
- [x] Use fallback XSD for when using the current documentation fails for some reason
- [ ] Implement CSV export
- [ ] Give the user an easy way to use the `mastr_table_to_db_model` returned from `Mastr.generate_data_model`. E.g. by adding a function that generates a Python code snippet with the SQLALchemy models/tables.
- [ ] Clear up the date situation. I made a couple of changes to `utils_download_bulk.py` because I found date handling unnecessarily complex. #697 changes the same code and adds support for retrieving available XML download dates. If #697 is merged, we have to update this retrieval logic to also retrieve the documentation download dates.
- [ ] Think about how we handle the transition from users' existing databases. Especially w.r.t. to translated databases and also all the renamed columns.
- [ ] Create usage examples
- [ ] Agree on a couple of open points and implement them
  - [ ] How to determine primary key of tables? By hardcoding it for MaStR tables we know? Or by checking the available columns and choosing the most likely one based on some hierarchy (e.g. "Id > MastrNummer > EinheitMastrNummer > …") [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/utils/sqlalchemy_tables.py#L12-L55)
  - [ ] How much do we want to adjust/normalize column names? [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/utils/xsd_tables.py#L16-L24)
  - [ ] Do we want to handle the case where adding only some columns to a table fails? [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/xml_download/utils_write_to_database.py#L200-L209)
- [ ] Go through the library and remove newly obsolete code

### Added
- Add the new method `Mastr.generate_data_model` that downloads the newest MaStR documentation and uses the XSD file to build SQLAlchemy models from the contained definitions

### Updated
- Update the method `Mastr.download` with two optional new arguments `mastr_table_to_db_table`, with which the user can pass their own database schema, and `alter_database_tables`, with which the user can prevent `open-mastr` from issuing any DDL statements.

### Removed
- Remove the method `Mastr.translate` because the user can now pass a dict mapping MaStR table names to DB tables. By adjusting their DB table names, they can achieve a similar result.

### Automation
Closes #516

### PR-Assignee

### Reviewer


FlorianK13:
@Simon-Will  The "deprecate" branch was merged and does not exist anymore, hence this PR should be directed to the "develop" branch.

PR #697, opened 2025-12-16
Add interactive download functionality for MaStR date selection #696

fharookshaik:
## Summary of the discussion

This documentation describes the implementation of interactive download functionality for the open-MaStR project. The enhancement addresses a key limitation where users could only download the latest MaStR export data (today's date) and had no visibility into available historical downloads.

Problem Solved
**Before**: Users were limited to downloading only today's MaStR export data through bulk method.
**After**: Users can now browse all available downloads and select any historical date

Key Features Implemented

**Interactive Date Selection**: Users can browse available downloads and select which date to download
**Programmatic Access**: Developers can fetch available dates and select downloads programmatically
**Backward Compatibility**: All existing functionality preserved - no breaking changes
**Enhanced User Experience**: User-friendly interfaces for both interactive and programmatic workflows

### Added

- Add interactive download functionality to browse and select MaStR export dates [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Add `get_available_download_links()` function to fetch available downloads from MaStR website [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Add `list_available_downloads()` function to display downloads in user-friendly format [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Add `select_download_date()` function for interactive date selection [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Add `browse_available_downloads()` method to Mastr class [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)

### Updated
- Enhance `download_xml_Mastr()` function with optional URL parameter for custom downloads [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Update `Mastr.download()` method with select_date_interactively parameter [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)
- Update `Mastr.download()` method docstring with comprehensive documentation for new parameter [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)

### Removed

- No functionality removed - maintains 100% backward compatibility [(#696)](https://github.com/OpenEnergyPlatform/open-MaStR/issues/696)

### Automation
Closes #696 

### PR-Assignee
- [ ] 🐙 Follow the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)
[...]

Issue #570, opened 2024-10-01
Get size of bulk download for loading bar component


FlorianK13:
Right now we have the total size of the download hardcoded. At the time of creating the bulk download code, BNetzA did not provide a total download size. This has changed now, so we could get the total size of the download from the requested url.

PR #718, opened 2026-01-24
516 dynamic tables and columns

### LINKED ISSUE #516: Dynamically add new tables to existing databases

### LINKED ISSUE #577: Apparent duplicates in `einheitenaenderungnetzbetreiberzuordnungen` are dropped
During import into `changed_dso_assignment` / `ChangedDSOAssignment` the log says
```
File 'EinheitenAenderungNetzbetreiberzuordnungen.xml' is parsed.
Data is cleansed.
778 entries already existed in the database.
```
However - if I understand the table right - the system operator may change multiple times so it's not a good idea to solely use `EinheitMastrNummer` as primary key:

https://github.com/OpenEnergyPlatform/open-MaStR/blob/6bc72541135c8eac66ed6565794e238f5e157957/open_mastr/utils/orm.py#L795-L807

To avoid losing this data, we should think of a primary key constraint on multiple columns such as
```
__table_args__ = (
        PrimaryKeyConstraint('EinheitMastrNummer', 'NetzbetreiberMastrNummerAlt', NetzbetreiberMastrNummerNeu),
)
```
However, there could be duplicates as well so maybe it's better to include the date like this:
```
__table_args__ = (
        PrimaryKeyConstraint('EinheitMastrNummer', 'Netzbetreiberzuordnungsaenderungsdatum'),
)
```

The table was added in #510

### LINKED ISSUE #661: Simplify `.to_csv` method
The .to_csv function is extremely complicated and full of artifacts from SOAP API times. We could instead:

- export the sqlite tables to csv, just as they are
- merge `_extended` and `_eeg` using pandas, which should roughly recreate the current behaviour

Simon-Will:
A first attempt at creating tables based on the MaStR XSD files.

Checklist until this is really ready:
- [x] Download current documentation and create database model from it
- [x] Use new database model for all the insertion code
- [x] Use fallback XSD for when using the current documentation fails for some reason
- [x] Implement CSV export
  - I radically simplified the existing CSV export. It was pretty complex, joined several tables and backfilled the basic units table. I found that a bit much for an export. There's probably a way to make it work in much the same way as it used to work, but I frankly didn't want to spend the time to fully understand all of what's going on there. Let's talk about it!
- [x] Implement translation feature
- [x] Give the user an easy way to use the `mastr_table_to_db_model` returned from `Mastr.generate_data_model`. E.g. by adding a function that generates a Python code snippet with the SQLALchemy models/tables.
  - I solved this by having `Mastr.generate_data_model` return SQLAlchemy core tables, not ORM models. They are easy to just print and a user can then copy them to their code & modify them. They are also the best common ground. After all, some users might not use the ORM.
  - There's also a function `format_mastr_table_to_db_table` that makes printing easy for the user.
- [x] Clear up the date situation. I made a couple of changes to `utils_download_bulk.py` because I found date handling unnecessarily complex. #697 changes the same code and adds support for retrieving available XML download dates. If #697 is merged, we have to update this retrieval logic to also retrieve the documentation download dates.
  - I reverted my changes regarding the dates.
  - I added the docs download to the download browsing, etc. Note that some old XSD files are invalid (e.g. 20240101) and cannot be read with XMLSchema. We fall back to the XSD files in the library in that case.
- [x] Think about how we handle the transition from users' existing databases. Especially w.r.t. to translated databases and also all the renamed columns.
  - My proposal: Since it is extremely difficult to provide an upgrade path from the old table & column names to the way they are done now, I think we should just tell users to adjust their existing queries so they fit the newest open-mastr version. This is fine imo because this whole thing here will trigger a major version bump anyway.
  - Please let's talk about table name translations. I'm using the old names [here in this new code](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718/changes#diff-8404e741de0e0d07088e2848c302972f56eda1d9b7a055d4ca68f7aebbb47c3aR198-R241), but would rather like to create new names that are closer to the names of the original MaStR export files.
- [x] Create usage examples
- [x] Address a couple of open points
  - [x] How to determine primary key of tables? By hardcoding it for MaStR tables we know? Or by checking the available columns and choosing the most likely one based on some hierarchy (e.g. "Id > MastrNummer > EinheitMastrNummer > …") [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/utils/sqlalchemy_tables.py#L12-L55)
    - I hard-coded the id column for the tables we know. For unknown future tables, a column "openMastrId" will be inserted. This is also done for the EinheitenAenderungNetzbetreiberzuordnungen table because there is no primary key.
  - [x] How much do we want to adjust/normalize column names? [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/utils/xsd_tables.py#L16-L24)
    - I decided to only do straightforward changes (MaStR -> MaStR, ß -> ss, deleting surrounding whitespace, etc.). No singularization/pluralization of column names à la VerknuepfteEinheitenMaStRNummern -> VerknuepfteEinheit.
  - [x] Do we want to handle the case where adding only some columns to a table fails? [Cf. this code](https://github.com/OpenEnergyPlatform/open-MaStR/blob/5625087a42a4994fbe4a53bd5f859a78ea06128d/open_mastr/xml_download/utils_write_to_database.py#L200-L209)
    - I decided not to add special handling for that.
- [x] Go through the library and remove newly obsolete code
- [x] Add tests

### Added
- Add the new method `Mastr.generate_data_model` that downloads the newest MaStR documentation and uses the XSD file to build SQLAlchemy models from the contained definitions

### Updated
- Update the method `Mastr.download` with two optional new arguments `mastr_table_to_db_table`, with which the user can pass their own database schema, and `alter_database_tables`, with which the user can prevent `open-mastr` from issuing any DDL statements.
- Change CSV export by removing joining tables. The tables are now exported as they are.
- Change the default names of tables and columns that are created and used for the import.

### Removed
- Remove the method `Mastr.translate`. The user can now get English table and column names by passing `english=True` to the `generate_data_model` or `download` method.

### Automation
Closes #516
Closes #577

### PR-Assignee

### Reviewer


pt-kkraemer:
On 19.02.2026 a new public version of the Gesamtdatenexport will have a [bugfix](https://www.marktstammdatenregister.de/MaStRHilfe/subpages/releasenotes.html) concerning attributes in the .xsd files:
grafik

This will probably have no effect on what you have done already right?

Simon-Will:
Interesting, thanks for pointing that out, @pt-kkraemer! It shouldn't make any difference for now because we make all attributes except the primary keys nullable anyway.

But I'll be sure to check out the difference in the XSD files to make sure I understand that point correctly.

As for this whole PR, it's almost ready now as you can see from the checklist. If anyone already has comments on the approach, please let them be heard. I don't think I will make substantial changes to the non-testing code anymore.

Simon-Will:
@FlorianK13, thanks for roughly going over the PR with me in the meeting! I updated the docs now and I spent some more time on the primary key thing. See https://github.com/OpenEnergyPlatform/open-MaStR/pull/718/changes/3fa4750f5bef5616c123e2a48ae8b2618e8c846a for how I did it now. Unfortunately, I don't see a way to use a composite primary key without any data loss for that table.

I can spend some time on creating views for the old table names on Monday.

I already set it to ready for review now, though, and invite you and the others to try it out. :) @pt-kkraemer, @nesnoj

open_mastr/utils/xsd_tables.py
```diff
@@ -0,0 +1,181 @@
...
+from open_mastr.utils.helpers import data_to_include_tables
+from open_mastr.utils.constants import COLUMN_TRANSLATIONS, TABLE_TRANSLATIONS
+
+_XML_SCHEMA_PREFIX = "{http://www.w3.org/2001/XMLSchema}"
+
+log = logging.getLogger("open-MaStR")
+
+# TODO: Should we really mess with the original column names?
+#  The BNetzA "choice" to sometimes write MaStR and sometimes Mastr is certainly confusing,
+#  but are we the ones who should change that?
```
FlorianK13: We should not be the ones to change that. But sadly we decided to do it anyway several years ago and now have to live with that decision 😮‍💨

open_mastr/utils/xsd_tables.py
```diff
@@ -0,0 +1,181 @@
...
+# Also TODO: Should we also apply the more opinionated normalization/renaming that is currently stored in orm.py?
+#  E.g. "VerknuepfteEinheitenMaStRNummern" -> "VerknuepfteEinheiten", "NetzanschlusspunkteMaStRNummern" -> "Netzanschlusspunkte", etc.
+def normalize_mastr_name(original_mastr_name: str) -> str:
+    # BNethA sometimes has MaStR, other times MaStR. We normalize that.
```
FlorianK13:
```suggestion
    """
    Replacing MaStR with Mastr. Also Replacing special German characters like ä -> ae
    """ 
```

open_mastr/utils/xsd_tables.py
```diff
@@ -0,0 +1,181 @@
...
+    # Also, in case the column names in the XSD contain äöüß, we replace them. This is probably a BNetzA oversight, but has happened at least once.
```
FlorianK13:
```suggestion

```

open_mastr/utils/xsd_tables.py
FlorianK13: I'd move some of those comments to docstrings, so that the function gets some explanation when hovering over it in your IDE. Not sure however if the docstring formatting works with these requested changes - the text highlighting in the github PR looks strange.

Simon-Will:
I just added some view functionality. Now I'm running the download & import with the current `develop` branch and will afterwards run the import again with the current state of this branch here to check if transitioning from an existing version will work. (This takes a while because the download takes a long time on my connection.)

Edit: Looks like it worked. Some column names of course change, but the views that preserve the old table names are in place.

FlorianK13:
> I just added some view functionality. Now I'm running the download & import with the current `develop` branch and will afterwards run the import again with the current state of this branch here to check if transitioning from an existing version will work. (This takes a while because the download takes a long time on my connection.)
> 
> Edit: Looks like it worked. Some column names of course change, but the views that preserve the old table names are in place.

So great that this View thing works :tada: I still found one bug, namely that the views now get the TempID 

image


Edit: This is not a bug in Views but it is for wind in general. Somehow in my local branch I have an additional commit from you called "Improve clarity around artificial primary keys" where the primary ID for wind is commented in `sqlalchemy_tables.py` - I have no idea where this is coming from, but it is not part of this PR so I guess you can ignore this error.

open_mastr/utils/xsd_tables.py
```diff
@@ -0,0 +1,181 @@
...
+            instance_name=main_element.name,
+            columns=columns,
+        )
+
+
+class InvalidXmlSchemaError(Exception):
+    pass
+
+
+def read_mastr_table_descriptions_from_xsd(
```
FlorianK13, line 145: If I see it correctly, there is no tests for this function yet. However if I understand it correctly a test would be very crucial, as we somehow need to test new fallback resources "Dokumatition-MaStR-...". In the future, with new releases we will update the fallback resource, but we have to be sure that this function still works with the updated zip file. Can you add a test for this purpose? Or is there already one that I did not see?

open_mastr/mastr.py
```diff
@@ -151,34 +239,64 @@ def download(
...
+            If set to True, prior downloaded MaStR zip files will be kept. Defaults to False.
+
+        mastr_table_to_db_table : Mapping from MaStR table name (str) to SQLALchemy Table, or None, optional
+            If given, downloaded data from a MaStR file will be stored in the SQLAlchemy table
+            associated with that file. The tables must exist already; they are not created.
+            Example: {"EinheitenWind": Table(...), "EinheitenSolar": Table(...), ...}
+
+            If None / not given, the mapping will be generated by calling
+            `Mastr.generate_data_model` (and thus downloading the MaStR documentation). In this
+            case, the tables will be created.
```
FlorianK13, line 296:
```suggestion
            If None / not given, the mapping will be generated by calling
            `Mastr.generate_data_model` (and thus downloading the MaStR documentation). You can use this function to define an own mapping. 
```

open_mastr/mastr.py
```diff
@@ -64,54 +66,140 @@ class Mastr:
...
+        metadata: Optional[MetaData] = None,
+        catalog_value_as_str: bool = True,
+        english: bool = False,
+    ) -> dict[str, Table]:
+        """Generate data model from MaStR documentation.
+
+        Download the MaStR documentation, extract the XSD files that describe the format of the
+        MaStR XML files, and generate SQLAlchemy tables from those XSD files. The tables are not
+        created in the database.
+
```
FlorianK13, line 118:
```suggestion
   You can use this method to create a mapping from the MaStR data model to your own data model. This mapping needs to be provided to the `Mastr.download` method.
    !!! example

        ```python
        from open_mastr import Mastr

        db = Mastr()
        mapping = db.generate_data_model()
        # edit your mapping here
        db.download(mastr_table_to_db_table=mapping)
        ```

```

FlorianK13: No idea if the indentation of this edit still matches with the needed indentation for correct docstring formatting.

open_mastr/mastr.py
```diff
@@ -64,54 +66,140 @@ class Mastr:
...
+        method: Literal["bulk"] = "bulk",
         data=None,
         date=None,
-        bulk_cleansing=True,
-        keep_old_downloads: bool = False,
         select_date_interactively: bool = False,
-        **kwargs,
+        bulk_cleansing: bool = True,
+        keep_old_downloads: bool = False,
+        mastr_table_to_db_table: Optional[Mapping[str, Table]] = None,
```
FlorianK13, line 218: I'm not so happy with the naming of this variable. For me it sounds like a boolean variable. Maybe `data_model_mapping` instead? Or `mastr_data_model_mapping`? Not sure however, variable naming is always subjective.

open_mastr/mastr.py
```diff
@@ -242,6 +381,7 @@ def download(
         delete_zip_file_if_corrupted(zipped_xml_file_path)
         delete_xml_files_not_from_given_date(zipped_xml_file_path, xml_folder_path)
 
+        # TODO: Why is this duplicated?
```
FlorianK13: No idea, let's just remove the print() statement.

open_mastr/__init__.py
```diff
@@ -3,4 +3,6 @@
 setup_project_home()
 
 # This import should be called after the setup project home, since .open_mastr folder must be defined first
+# TODO: Why must we create the folder before importing??
```
FlorianK13: I think the reason is that both the Mastr class and the open_mastr.soap_api.download.MaStRAPI need the project folder. But there could also be a method that creates the folder in the init of these two classes.

However I'd say that it works as it is right now and propose that we just leave the setup_project_home here, or what do you think?

FlorianK13 (commented):
Some more comments

open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -339,29 +342,16 @@ def full_download_without_unzip_http(
...
-            )
-        else:
-            print(f"MaStR file already present: {save_path}")
-            return None
-
     warning_message = (
         "Warning: The servers from MaStR restrict the download speed."
         " You may want to download it another time."
     )
+    # TODO: Explain this number
```
FlorianK13: See #570

open_mastr/utils/xsd_tables.py
```diff
@@ -0,0 +1,181 @@
...
+    translated = COLUMN_TRANSLATIONS.get(normalized_mastr_column_name)
+    if not translated:
+        log.warning(f"No translation available for column {normalized_mastr_column_name!r}")
+    return translated
+
+
+def translate_mastr_table_name(normalized_mastr_table_name: str) -> Optional[str]:
+    translated = TABLE_TRANSLATIONS.get(normalized_mastr_table_name)
+    if not translated:
+        log.warning(f"No translation available for column {normalized_mastr_table_name!r}")
```
FlorianK13:
```suggestion
        log.warning(f"No translation available for table {normalized_mastr_table_name!r}")
```

open_mastr/mastr.py
```diff
@@ -254,151 +394,52 @@ def download(
...
-                ),
-                data_table=addit_table,
-                chunksize=chunksize,
-            )
-
-        # FIXME: Currently metadata is only created for technology data, Fix in #386
-        # Configure and save data package metadata file along with data
-        # save_metadata(data=technologies_to_export, engine=self.engine)
-
-    def translate(self) -> None:
```
FlorianK13, line 332: I would not remove `translate`, instead it should have no code anymore and raise an error saying that the functionality is moved to `Mastr.download` with parameter `english`

FlorianK13: Also some docstring with the same content (deprecated, moved to .download method) would be great.

open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -537,15 +567,93 @@ def select_download_date():
...
+    newest_url = "https://www.marktstammdatenregister.de/MaStRHilfe/files/gesamtdatenexport/Dokumentation%20MaStR%20Gesamtdatenexport.zip"
+    if url:
+        preferred_url = url
+        fallback_url = None
+    else:
+        if bulk_date_string:
+            dt = datetime.strptime(bulk_date_string, "%Y%m%d")
+            stichtag_url = (
+                "https://download.marktstammdatenregister.de/Stichtag/"
+                "Dokumentation%20MaStR%20Gesamdatenexport%20"
```
pt-kkraemer: Missing "t" in "Gesamtdatenexport" or? I could not find a link with this particular error on the MaStR site

open_mastr/xml_download/utils_download_bulk.py
```diff
@@ -223,26 +226,40 @@ def _parse_date_string(date_str):
...
+    """Checks if an existing download contains the xml-files corresponding to the bulk_data_list."""
+    if os.path.exists(save_path):
+        if not _find_missing_files(save_path, bulk_data_list):
+            log.info(
+                "MaStR XML ZIP file already present and has all info. Not downloading again."
+                f" Existing file: {save_path}"
+            )
+            return True
+        log.info(
+            f"MaStR XML ZIP file already present but missing the following data: {bulk_data_list}"
```
pt-kkraemer: Comprehension question: what happens if only a part of the `bulk_data_list` is already downloaded? Would we not show here that the whole `bulk_data_list`  is missing? Could be misleading.

Edit: I tested the following case where I first downloaded `"biomass"` and then `["biomass","wind"]` and in the second run `"biomass"` got downloaded again:
grafik

CITATION.cff
```diff
@@ -38,6 +38,9 @@ authors:
     given-names: "Kevin"
     alias: "pt-kkraemer"
     affiliation: "ProjectTogether gGmbH"  
+  - family-names: "Will"
+    given-names": "Simon"
```
pt-kkraemer: Random " after `given-names`

CITATION.cff
```diff
@@ -38,6 +38,9 @@ authors:
     given-names: "Kevin"
     alias: "pt-kkraemer"
```
pt-kkraemer: Missing "@" in front of my alias

open_mastr/utils/xsd_tables.py
Simon-Will: Moved to docstring.

open_mastr/utils/xsd_tables.py
Simon-Will: I would like to keep this line because it explains that this äöüß is based on an actual error that BnetzA made. Why would you like to remove it?

open_mastr/mastr.py
Simon-Will: Used your suggestion with slight change in wording.

open_mastr/mastr.py
Simon-Will: Thanks, nice! I used this.

open_mastr/mastr.py
Simon-Will: I almost always call my dictionaries names like `x_to_y`.

In my opinion, having `_mapping` is redundant in a similar way as hungarian notation is redundant. It should be apparent from the variable's type that it's a mapping. What's more interesting is what is mapped to what. I think that's less clear in `data_model_mapping`.

I agree that `mastr_table_to_db_table` is a bit verbose. Since I always use names like this, I'm used to its structure and prefer it. But we can also rename it if you like.

open_mastr/mastr.py
Simon-Will: Removed

open_mastr/__init__.py
Simon-Will: Yes, no point in changing this now. I removed the TODO comment. Thanks for the explanation!

open_mastr/xml_download/utils_download_bulk.py
Simon-Will: Thanks! I added a comment that links to this issue.

open_mastr/utils/xsd_tables.py
Simon-Will: Thanks, good catch.

open_mastr/mastr.py
Simon-Will: Makes sense! I added a `translate` method stub.

open_mastr/xml_download/utils_download_bulk.py
Simon-Will: Ooh, thanks a lot for catching that!

open_mastr/xml_download/utils_download_bulk.py
Simon-Will: True! I think I missed the incremental nature of the download when refactoring this part of the code a bit. 😬 

I now changed it up a bit more and it should be correct and incremental now.

CITATION.cff
Simon-Will: Thanks, removed.

CITATION.cff
Simon-Will: Added an @

open_mastr/utils/xsd_tables.py
Simon-Will: This function is tested as part of the tests in `test_mastr.py`. But you're right: We should also add a test for the fallback to the stored docs. I just did that, calling it `test_mastr_generate_data_model_fallback_to_included_docs`.

Simon-Will:
Hi @FlorianK13 and @pt-kkraemer, thanks a lot for your reviews and suggestions! I addressed your comments with the two latest commits. Could you have another look?

open_mastr/mastr.py
```diff
@@ -242,163 +398,58 @@ def download(
         delete_zip_file_if_corrupted(zipped_xml_file_path)
```
FlorianK13: delete_zip_file_if_corrupted and delete_xml_files_not_from_given_date are duplicated from above. This might have happend in some earlier merge. However they should be called only once - these two calls need to be deleted.

Simon-Will: Oh, true! But apparently this error is already present in the `develop` branch for quite a while. I'd say let's fix it there. After all, we do want to release a new version before the release that includes this PR, right?

pt-kkraemer:
I just had the same error as #723 when using the latest pull from your fork @Simon-Will.
grafik

@FlorianK13 wrote in #723 that this will be fixed, I don't think it does. Maybe we can talk about this in our next dev-meeting?

tests/test_mastr.py
```diff
@@ -1,110 +1,336 @@
...
+        expected_url,
+        body=outer_zip_content.getvalue(),
+        content_type='application/zip'
+    )
+
+    mastr_table_to_db_table = mastr.generate_data_model(
+        date="20260301"
+    )
+    # We expect that reading the invalid XSD Netze.xsd fails, triggering the fallback.
+    assert len(caplog.messages) == 1
```
FlorianK13, line 310: This assertion fails when I run the tests, with len(caplog.messages)=0

open_mastr/mastr.py
FlorianK13: I'm not sure if we want to do this. A release takes up several hours per developer, and if we stick to our rule that at least two persons check the codebase before releasing this takes almost a day in total. Sure we should do hotfixes for stuff like this, but all that takes just so much time. 

We run more and more into this problem that we (or at least I) do not have the time resources to stick to all our best practices. Or if we do, everything takes ages and we do not manage any fast releases, code reviews etc anymore. I think we should discuss this topic in our next meeting.

open_mastr/utils/xsd_tables.py
FlorianK13: I merged both into the docstring. But we can also keep this as a comment.

FlorianK13 (changes requested):

tests/test_mastr.py
Simon-Will: It did fail indeed! Not sure if it had ever worked for me before, but the reason it failed was that `caplog` couldn't capture our custom `open-MaStR` logger's logs. `caplog` (by default) attaches to the root logger, but the logs never reached the root logger because the `open-MaStR` logger we use was defined with `propagate: no` in `logging.yml`.

I now changed `logging.yml` so that the `open-MaStR` logger propagates its logs to the root logger and attached the file and the console handler to the root logger instead. This not only solves this log capturing problem, I also think that it makes sense because attaching handlers at the root is more useful in general. After all, an application's logs should usually all go to the same place, regardless of what part they are logged from. **For the change to take effect locally, you must delete your `~/.open-MaStR/config/logging.yml` so that it's generated again.**

(In general, I would prefer not having this `logging.yml`, at all, because I think of `open-MaStR` primarily as a library. I would like to leave configuration of logging to the user. But not the time & place to discuss this.)

open_mastr/mastr.py
Simon-Will: We resolved on Thursday to indeed do a release before merging here.

Simon-Will:
All discussions resolved. I'm waiting for the 0.17 release and will then update this branch before it can be merged. :)

FlorianK13:
@Simon-Will can you resolve the conflicts? Afterwards I think we can merge.

Simon-Will:
@FlorianK13, conflicts resolved. I'm running the download & import right now to see if it goes well.

Simon-Will:
The import is going well. So this is ready again, but I think we want to do #736 first.

Simon-Will:
I merged develop into this pull request branch again. So it's ready for merge again. To whoever wants to merge: I'd recommend squashing all these commits I made. The history isn't really clean enough to be useful imo.

FlorianK13 (dismissed):

CHANGELOG.md
```diff
@@ -8,8 +8,14 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
 
 ## [v0.xx.x] Unreleased - 202x-xx-xx
 ### Added
+- Add the option to pass a custom database schema
```
FlorianK13: We discussed in a call and @nesnoj pointed out if we can use a LLM to generate an extended changelog - since a lot of stuff happend here. I did this and it created this new changelog:

## [v0.xx.x] Unreleased - 202x-xx-xx                                                                                                                                                          
  ### Added                                                                                                                                                                                     
  - Add `Mastr.generate_data_model` method that downloads the current MaStR documentation and generates SQLAlchemy tables from the XSD definitions; supports `english=True` for English column  
  names                                                                                                                                                                                         
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
  - Add `mastr_table_to_db_table` argument to `Mastr.download` to pass a custom database schema                                                                                                 
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
  - Add `alter_database_tables` argument to `Mastr.download` to prevent open-mastr from issuing DDL statements
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
                                                                                                                                                                                                
  ### Changed                                                                                                                                                                                   
  - Switch to dynamic table generation based on parsing of XSD files from the MaStR documentation; fall back to bundled XSD files if the downloaded documentation is invalid                    
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
  - Change default table and column names to align more closely with the original MaStR export file names                                                                                       
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
  - Simplify CSV export by removing cross-table joins; tables are exported as-is                                                                                                                
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)                                                                                                                           
                                                                                                                                                                                                
  ### Removed                                                                                                                                                                                   
  - Remove `Mastr.translate`; English table and column names are now available via the `english=True` parameter in `generate_data_model` and `download`
    [#718](https://github.com/OpenEnergyPlatform/open-MaStR/pull/718)

FlorianK13 (changes requested):
@Simon-Will can look at my comment and do this change to the CHANGELOG if you agree with the described changes there? Afterwards I an squash commit merge the PR.

CHANGELOG.md
Simon-Will: I'm so sorry! I missed this completely and somehow thought the PR had been merged already.

Thanks for this nice CHANGELOG! I used it.

FlorianK13 (approved):

CHANGELOG.md
FlorianK13: No problem. It is finally merged :fireworks:

PR #749, opened 2026-05-22
670 improve testing

### LINKED ISSUE #670: Streamline unit and integration tests
The tests should be improved to enhance maintainability, coverage, and performance:

**High Priority**

- [x] Test Isolation: Tests use real user home directories (`~/.open-MaStR`) and modify actual data files instead of using temporary directories
- [x] Duplicate Fixtures: `zipped_xml_file_path` fixture defined in 3 different files without cleanup. Move to `conftest.py`?
- [ ] Bug: Function Name Typo: text_extract_sql_table_name should be test_extract_sql_table_name (test_utils_write_to_database.py:89)
- [ ] Missing Test Configuration: No pytest.ini for centralized test settings and coverage configuration

**Medium Priority**

- [ ] Magic Numbers: Hard-coded values like 10000, 1000000, credential lengths without named constants
- [ ] Poor Assertions: Confusing boolean comparisons (assert False == df.empty) and missing error messages
- [ ] No Mocking: Real API calls in tests make them slow and dependent on external services
- [ ] Repetitive Tests: test_gen_url() has 12 nearly identical assertions instead of using @pytest.mark.parametrize

**Low Priority**

- [ ] Missing Docstrings: Tests lack explanations of what they verify and why
- [ ] Test Coverage Gaps: No tests for error handling, edge cases, or concurrent operations
- [ ] Inconsistent Naming: Mix of test_Mastr_init vs test_mastr_download conventions

FlorianK13:
# General
This is a first step for making our tests better. It creates a mockup for the bulk download and it skips the API tests if no credentials are provided (which hopefully makes the CI run successfully, also for PRs from outside)

# SUmmary of changes
- **Add `tests/data/Gesamtdatenexport_mockup.zip`**: a small (~120 KB) mockup of the MaStR bulk download, generated by `scripts/create_bulk_download_mockup.py`. Contains 100 rows per XML table for wind, biomass, hydro

- **Refactor `tests/conftest.py`**: replace the scan for existing local ZIP files with two fixtures (`mockup_xml_zip_in_output_dir`, `mockup_docs_zip_in_output_dir`) that copy the bundled files into a temporary directory. The `output_dir` fixture now sets `MASTR_PROJECT_HOME_DIR` via `monkeypatch.setenv` so the `Mastr()` instance automatically picks up the temp directory without needing an explicit `output_dir` argument.

- **Refactor `tests/test_mastr.py`** — remove all `@pytest.mark.skipif` guards that depended on real data and split the single large download integration test smaller ones

- Switch `test_mastr_generate_data_model_fallback_to_included_docs` from `caplog` to `patch.object` on the logger as somehow the caplog thing always failed

- **Refactor `tests/soap_api/test_download.py`** — add a `requires_real_credentials` marker (using a runtime credential check) so that tests requiring real MaStR credentials are explicitly skipped when credentials are unavailable -> **THIS HOPEFULLY FIXES THE GH CREDENTIAL ISSUE WHERE CI FAILED FOR OUTSIDE PRs**
- **Delete `tests/test_credentials.py`** — the credential test required real secrets and provided no value in CI.
- Since pandas>3.0 it handles dtypes differently - i needed to adapt some tests so that mismatching dtypes do not cause failure anymore

### PR-Assignee

### Reviewer


tests/soap_api/test_download.py
```diff
@@ -1,5 +1,21 @@
...
+import responses
+
+
+def _check_real_credentials():
+    try:
+        api = MaStRAPI()
+        result = api.GetAktuellerStandTageskontingent()
+        return result.get("Ergebniscode") == "OK"
+    except Exception:
+        return False
```
Simon-Will, line 12: I find `except Exception` too broad here. If there are stored credentials that are supposed to be correct, but something in the construction of `MaStRAPI` or the `GetAktuellerStandTageskontingent` call fails, then two of the three tests are skipped, even though we probably rather want them to show some failure.

What do you think about raising a new `MissingMaStRAPIKeyError` at the end of `MaStRAPI.__init__` if `self._key` is still `None there? Then we could rename this function here from `_check_real_credentials` to `_check_credentials_configured` and catch the `MissingMaStRAPIKeyError` instead.

tests/xml_download/test_utils_cleansing_bulk.py
```diff
@@ -43,37 +27,36 @@ def test_cleanse_bulk_data(existing_xml_zip_in_output_dir: Path) -> None:
...
+) -> None:
     katalogwerte = create_katalogwerte_from_bulk_download(
-        zipped_xml_file_path=existing_xml_zip_in_output_dir
+        zipped_xml_file_path=mockup_xml_zip_in_output_dir
     )
-    assert type(katalogwerte) == dict
-    assert len(katalogwerte) > 1000
-    assert type(list(katalogwerte.keys())[0]) == int
+    assert isinstance(katalogwerte, dict)
+    assert len(katalogwerte) > 0
```
Simon-Will: What's the reason for changing this assertion from `> 1000` to `> 0`? The mockup Katalogwerte should still be untrimmed, right?

tests/test_mastr.py
```diff
@@ -1,168 +1,187 @@
...
-    assert len(df_wind) > 10000
-    assert len(df_biomass) > 10000
+    df_hydro = pd.read_sql("EinheitenWasser", con=mastr.engine)
+    assert 0 < len(df_biomass) <= NUMBER_ROWS_IN_MOCK_XML_FILES
+    assert 0 < len(df_hydro) <= NUMBER_ROWS_IN_MOCK_XML_FILES
 
-    # Test that we can pass a list of data.
-    mastr.download(data=["wind", "nuclear"])
+
+def test_download_accumulates_across_calls(
```
Simon-Will, line 114: I like these test names! :)

Simon-Will (changes requested):
Thanks a lot. It's looking quite good, I think! I only have some minor comments on most changes and one bigger thought:

I'm thinking a bit about how to trim the XML files and which files to choose for testing on. The current tests always work with wind, biomass, water and nuclear since these are small files and can be imported in a reasonable time for a test. I want to make two independent points:
* With trimmed files, we can also in a short time import multiple files because they only contain 100 records each. This way, we would also test the multi-file imports, which is currently a blind spot in the tests.
* The trimming works by using the first 100 records in a file. I think these are also the oldest records in the MaStR database, so they are quite a biased sample. It could happen that newer records contain some data that cannot be imported, but this would never be tested. This could be alleviated by choosing a random sample in the trimming process or by using solar so that we automatically have newer records by always having the first 100 records in each file.
** To be honest, this point is a bit moot since we already don't test the majority of the file types. If there were some data in some of them that couldn't be imported, the tests wouldn't catch it anyway.

What do you think about these?

tests/xml_download/test_utils_cleansing_bulk.py
FlorianK13: Yes you are right, I'll change it.

FlorianK13:
> With trimmed files, we can also in a short time import multiple files because they only contain 100 records each. This way, we would also test the multi-file imports, which is currently a blind spot in the tests.

I am not 100% sure what you mean here. Is your suggestion to also include large tables like solar, storages, ...? If yes I think I agree. The script will take a lot longer to run, but we only need to run it once in a while. And it has the advantage that more (all) technologies are part of the tests.

Edit: Now I understand what you mean with multi-file: This is about technologies that have multiple xml files, like EinheitenSolar_1, _2, ... I will think of a way, maybe have maximum of three files per technology.

tests/soap_api/test_download.py
FlorianK13: Sounds good, but I guess this should be part of a new issue then.

FlorianK13: See #750

FlorianK13:
@Simon-Will can you re-review? I resolved your requested changes and did the following:
- The bulk download mockup now includes all tables from the db.download() docs
- If zipped downloads consist of **one** file, it will be written to **one** mockup file with 100 entries
- If zipped downloads consist of **two or more** files, they will be written to **exactly four** mockup file with 25 entries. The algorithm puts roughly an equal amount of entries from each xml file into the final mockup file. 
  - at the end only some entries are thrown away to make sure that each mockup has at most 100 entries
- I added a few more data tests, but then stopped as rigorous testing would be part of the next PR

Simon-Will:
@FlorianK13, hi, I was on vacation last week, but now took a look. I'm a bit confused since I cannot see new commits, the `> 0` assertion is still there and the mockup seems like it hasn't changed. Have you maybe forgotten to push the changes? Or am I missing something?

FlorianK13:
yeah sorry, forgot to push. Now the branch is ready for your review @Simon-Will

Simon-Will (approved):
I like the new creation of the mockup files. Very nice! Looks good to me!

Issue #751, opened 2026-06-04
Increase FAIRness of open-mastr


FlorianK13:
We could apply here with deadline 19.07.2026: https://fdm-nds.de/index.php/fair_award_2026/ 

I will go through the fairness checker for software from their homepage and collect everything we need to do here.

FlorianK13:
- [x] Connect OpenEnergyPlatform/open-MaStR to zenodo - like [here](https://doi.org/10.5281/zenodo.17775947)
- [x] Make one release so that the mirroring to zenodo starts
- [x] Create a code meta file with https://autocodemeta.linkeddata.es/ as soon as a doi on zenodo is available
- [x] Add zenodo doi to citation.cff

nesnoj:
If we can't make it to resolve the current bugs before the 19th, would it be an option to make a minor release with only the changes required for application?

FlorianK13:
@nesnoj Since it is a release with new features I would still go to 0.18.0 and then resolve bugs in a 0.18.1 or so

FlorianK13:
@nesnoj @Simon-Will I'll do the release on July 16th and I will also do the 0.18.1 release when all fixes are merged.

nesnoj:
> [@nesnoj](https://github.com/nesnoj) [@Simon-Will](https://github.com/Simon-Will) I'll do the release on July 16th and I will also do the 0.18.1 release when all fixes are merged.

Sounds good, thx!

Simon-Will:
I wanted to understand better what we have to do here. So I went through [Checklist_FAIR_Software_v01.pdf](https://fdm-nds.de/wp-content/uploads/2026/06/Checklist_FAIR_Software_v01.pdf) and tried to answer the questions. Here is how I see it. @FlorianK13, do you agree?

1. Findable
- [x] Can the software be found via search engines?
  - Within the first couple of hits in **DuckDuckGo**:
    - ✅ "open-mastr"
    - ✅ "download mastr data"
    - ✅ "download marktstammdatenregister"
    - ❌ "marktstammdatenregister herunterladen"
      - [https://github.com/marktstammdatenregister-dev/mastr](https://github.com/marktstammdatenregister-dev/mastr) is found! That looks unmaintained, so I asked to maybe link to open-MaStR: https://github.com/marktstammdatenregister-dev/mastr/issues/57
  - Within the first couple of hits in **Google** (could be improved!):
    - ✅ "open-mastr"
[...]

PR #753, opened 2026-06-09
Add more test cases for `mastr.download()`

### LINKED ISSUE #670: Streamline unit and integration tests

FlorianK13:
Continues the work on #670


FlorianK13:
Hi @Simon-Will and @kvnkrmr - could one of you review these changes. With regards to #751 I would like to release v0.18.0 soon. Do you see anything that needs to be done before a release?

kvnkrmr:
I can review this week!

kvnkrmr:
If we use python>=3.12 for testing we should also change the [pyproject.toml](https://github.com/OpenEnergyPlatform/open-MaStR/blob/670-add-more-tests/pyproject.toml) to require python>=3.12?

kvnkrmr (approved):
Looks good to me, new tests pass

FlorianK13:
> If we use python>=3.12 for testing we should also change the [pyproject.toml](https://github.com/OpenEnergyPlatform/open-MaStR/blob/670-add-more-tests/pyproject.toml) to require python>=3.12?

I found on https://packaging.python.org/en/latest/guides/dropping-older-python-versions/ 

> Metadata 1.2+ installers, such as Pip, will adhere to this specification by matching the current Python runtime and comparing it with the required version in the package metadata. If they do not match, it will attempt to install the last package distribution that supported that Python runtime.

@kvnkrmr So I would not change the `requires-python` to 3.12 as users with python < 3.12 will then only be able to install open-mastr < 0.18.0 and I think there are no reasons why open-mastr doesnt work with lower python versions, we just do not test it anymore.

kvnkrmr:
Ok, yes makes sense

PR #755, opened 2026-06-09
Remove dead code :skull:

### LINKED ISSUE #591: Use a python profiler to find unused code
We can use a python profiler, [like here](https://docs.python.org/3/library/profile.html) to find functions and classes in open-mastr that are outdated and not used anymore.

FlorianK13:
Used `vulture` to find dead code.


FlorianK13:
There is a whole lot of code related to `metadata` that is dead. What is the status here? Do we know if the metadata creation for zenodo still works? When did we test it the last time? Is it still up to date (I doubt that since it is based on hardcoded .json files that somebody wrote ages ago).

PR #761, opened 2026-07-06
robust parsing of xsd metdatadata files

### LINKED ISSUE #760: `xsd.zip` not a zip anymore?
> xsd.zip not a zip anymore (but a directory) in my MaStR download from today, e.g. https://github.com/OpenEnergyPlatform/open-MaStR/blob/develop/open_mastr/utils/xsd_tables.py#L174 

 _Originally posted by @IsGut in [#758](https://github.com/OpenEnergyPlatform/open-MaStR/issues/758#issuecomment-4874266587)_

I downloaded the metadata files from marktstammdatenregister.de again. If I understand correctly we have this issue:

## Past folder structure
```sh
- Dokumentation MaStR Gesamtdatenexport.zip
  - xsd.zip
```
## Todays folder structure
```sh
- Dokumentation MaStR Gesamtdatenexport.zip
  - xsd (no zip)
```

FlorianK13:
See #760 

Idea is to allow both for xsd.zip and unzipped xsd to be parsed.

### PR-Assignee

### Reviewer

### DIFF
--- a/open_mastr/utils/xsd_tables.py
+++ b/open_mastr/utils/xsd_tables.py
@@ -3,7 +3,7 @@ import os
 from enum import auto, Enum
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional, Union, Iterator
+from typing import Optional, Union, Iterator, IO
 from zipfile import ZipFile
 import xmlschema
 from xmlschema.validators.simple_types import XsdAtomicBuiltin, XsdAtomicRestriction
--- a/open_mastr/utils/xsd_tables.py
+++ b/open_mastr/utils/xsd_tables.py
@@ -162,7 +162,9 @@ class InvalidXmlSchemaError(Exception):
 
 
 def _iterate_xsd_files(source: Union[Path, str]) -> Iterator[tuple[str, IO[bytes]]]:
-    """Iterate over .xsd files in either a zip containing xsd or xsd.zip.
+    """Iterate over .xsd files in a ZIP file or directory.
+    
+    Source can either be a directory directly containing `*.xsd` files, or a ZIP file that contains either an `xsd` directory or another ZIP file called `xsd.zip`.
 
     Yields tuples of (name, file_object) where file_object is a context manager.
     """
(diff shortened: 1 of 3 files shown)


open_mastr/utils/xsd_tables.py
```diff
@@ -142,40 +161,64 @@ class InvalidXmlSchemaError(Exception):
     pass
 
 
+def _iterate_xsd_files(source: Union[Path, str]):
```
Simon-Will: If you want, you can add a return type annotation:

```suggestion
def _iterate_xsd_files(source: Union[Path, str]) -> Iterator[tuple[str, IO[bytes]]]:
```

open_mastr/utils/xsd_tables.py
```diff
@@ -142,40 +161,64 @@ class InvalidXmlSchemaError(Exception):
...
+    """Iterate over .xsd files in either a zip containing xsd or xsd.zip.
```
Simon-Will: I think this could be a bit clearer:

```suggestion
    """Iterate over .xsd files in a ZIP file or directory.
    
    Source can either be a directory directly containing `*.xsd` files, or a ZIP file that contains either an `xsd` directory or another ZIP file called `xsd.zip`.
```

open_mastr/utils/xsd_tables.py
```diff
@@ -142,40 +161,64 @@ class InvalidXmlSchemaError(Exception):
...
+
+    Yields tuples of (name, file_object) where file_object is a context manager.
+    """
+    source_path = Path(source)
+
+    if source_path.is_dir():
+        # Case: source is already an unzipped directory containing .xsd files
+        for xsd_file in source_path.glob("**/*.xsd"):
```
Simon-Will, line 175: Perfectly fine and clear like this, but just fyi: You can also use `rglob` for this instead of `**/`.

Simon-Will (dismissed):
Thanks for taking this on! The logic makes sense to me and it looks sound! I only have some minor comments that you can also ignore.

One small change to how it was previously is that directories ending in `.xsd` were ignored whereas now they would be treated like an XSD file and would probably lead to an error. However, it's extremely likely that this will never happen, so I think it's fine!

nesnoj:
Tested and works for today's unzipped xsd

Simon-Will (approved):

PR #765, opened 2026-07-08
Race condition cause silent data loss

### LINKED ISSUE #763: Race condition when importing chunked XML files in parallel can cause silent data loss
When bulk XML imports are executed in parallel using ProcessPoolExecutor, there appears to be a race condition that can lead to silent data loss.

Each worker calls process_xml_file(), which checks whether it is processing the first chunk of a dataset:

if is_first_file(file_name):
    delete_all_existing_rows(db_table=db_table, engine=engine)

The assumption seems to be that the worker processing *_1.xml will always execute before workers processing later chunks. However, ProcessPoolExecutor does not guarantee execution order.

Possible execution sequence
Worker B starts processing AnlagenEegSolar_2.xml.
Worker B inserts and commits its rows.
Worker A (scheduled later) starts processing AnlagenEegSolar_1.xml.
is_first_file() returns True.
delete_all_existing_rows() is executed, deleting both the old data and the rows that Worker B has just inserted.
Worker A inserts only the contents of AnlagenEegSolar_1.xml.

The import completes successfully, but data from previously processed chunks is lost without any error being reported.

Why this happens

The table cleanup is performed inside each worker process rather than once before parallel processing begins. Since worker execution order is non-deterministic, there is no guarantee that the deletion occurs before any inserts for the same table.

Suggested fix

Perform table cleanup once in the parent process before creating the ProcessPoolExecutor (or otherwise synchronize the deletion), so that workers only perform inserts.

## Workflow checklist
- [ ] I have checked [the documentation](https://open-mastr.readthedocs.io/en/latest/) and confirmed this issue is not already addressed there.
- [ ] I am aware of the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)

FlorianK13:
See #763 

### PR-Assignee

### Reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
 
 
   ### Changed
+  - Fix race condition in parallel bulk XML import that resulted in silent data loss
+    [#765](https://github.com/OpenEnergyPlatform/open-MaStR/pull/765)
   - Refactor data path configuration: remove `get_data_version_dir()`, add `MASTR_PROJECT_HOME_DIR` env var support to `get_project_home_dir()`
     [#748](https://github.com/OpenEnergyPlatform/open-MaStR/pull/748)
   - Switch to dynamic table generation based on parsing of XSD files from the MaStR documentation; fall back to bundled XSD files if the downloaded documentation is invalid
--- a/open_mastr/xml_download/utils_write_to_database.py
+++ b/open_mastr/xml_download/utils_write_to_database.py
@@ -75,6 +75,11 @@ def write_mastr_xml_to_database(
                 )
             )
 
+    # Clear each target table exactly once in the parent process
+    db_tables_to_clear = {db_table.name: db_table for _, db_table, *_ in threads_data}
+    for db_table in db_tables_to_clear.values():
+        delete_all_existing_rows(db_table=db_table, engine=engine)
+
     interleaved_files = interleave_files(threads_data)
     number_of_processes = get_number_of_processes()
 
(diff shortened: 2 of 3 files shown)


HassanRady (approved):
Thanks for the quick turnaround! I reviewed the PR, and the changes look good to me.

nesnoj:
mastr is currently under maintenance and not testable.
I can give it a try later today.

Simon-Will (dismissed):
Nice and clear solution. Thanks, @FlorianK13! And thanks for noticing, @HassanRady; good to read you here! :)

I also ran the code of this branch (merged with #761 to get the newest xsd) and it worked well: The tables I wanted to download were deleted and reimported, but existing other tables & rows that I didn't want to download stayed untouched.

nesnoj:
In my today's run, `EinheitenSolar` is empty, trying to figure out if this relates to the multiprocessing.

(same condition as @Simon-Will so branch `763-race-condition-cause-silent-data-loss` with merged `760-xsdzip-not-a-zip-anymore`.)

nesnoj:
> In my today's run, `EinheitenSolar` is empty, trying to figure out if this relates to the multiprocessing.
> 
> (same condition as @Simon-Will so branch `763-race-condition-cause-silent-data-loss` with merged `760-xsdzip-not-a-zip-anymore`.)

It's not related to mp, so it's fine regarding this PR.
I 'll create a new issue.

FlorianK13:
Needs new review after I solved merge issues in the Changelog :| I'd bypass the rule and merge anyway.

Simon-Will:
@FlorianK13, I think in cases like this it makes total sense to not wait for another review, especially with how long I (and maybe others) often take to get around to it. Thanks for merging!

FlorianK13:
@Simon-Will thought the same then, those github rules are nice but some flexibility in the process is also fine :)

PR #780, opened 2026-07-17
751 increase fairness open mastr

### LINKED ISSUE #751: Increase FAIRness of open-mastr

FlorianK13:


codemeta.json
```diff
@@ -0,0 +1,198 @@
...
+        "python",
+        "solar",
+        "Markstammdatenregister",
+        "Energy",
+        "Dataset",
+        "Solar",
+        "Wind",
+        "Energy-data",
+        "OEP",
+        "OpenEnergyFamily"
```
Simon-Will: How where these generated? I guess they are case-sensitive? Then I'd add "Python", "wind", "BNetzA", and other case variants.

Also, I propose adding "bundesnetzagentur", "Bundesnetzagentur", "MaStR" and "mastr".

The CodeMeta specs don't say anything about case-sensitivity, as far as I can see. But in practice, everyone is probably going to lower-case these before performing any search.

FlorianK13: The file is generated from https://autocodemeta.linkeddata.es/

I think the metadata is taken from the pyproject.toml file. Hence I will add these new keywords to the pyproject toml file as well.

FlorianK13:
@Simon-Will  I'll merge it now as these changes are relevant for the FAIR competition. If changes are needed I can work on this in a second PR.

PR #790, opened 2026-08-20
Fix table name lookup in interleave_files #789

### LINKED ISSUE #789: Fix table name lookup in interleave_files to enable parallel worker interleaving
## Description of the issue

When building threads_data in write_mastr_xml_to_database, index 2 is the database connection URL string (str(engine.url)), while index 1 is the Table object.

In interleave_files, table_name = item[2] groups all files under the database URL rather than their table name, resulting in a single group and causing no interleaving to occur. This leads to parallel workers processing the same table simultaneously and colliding on database locks.

```Python
threads_data.append(
                (
                    file_name,
                    db_table,
                    str(engine.url),  # index 2
                    engine.url.password,
                    zipped_xml_file_path,
                    bulk_download_date,
                    bulk_cleansing,
                    alter_database_tables,
                )
            )

....
 
for item in threads_data:
        table_name = item[2]
        if table_name not in files_grouped_by_table:
            files_grouped_by_table[table_name] = []
        files_grouped_by_table[table_name].append(item)
```

Modifying the lookup to item[1] (item[1].name if hasattr(item[1], "name") else str(item[1])) correctly restores round-robin file interleaving.

## Workflow checklist
- [x] I have checked [the documentation](https://open-mastr.readthedocs.io/en/latest/) and confirmed this issue is not already addressed there.

HassanRady:
### Summary of Changes
- Fixed `interleave_files` to extract table name from `item[1]` (`Table` / string) instead of `item[2]` (database connection URL).
- Added comprehensive unit tests covering both string names and SQLAlchemy `Table` instances with realistic database URLs.
- Added author details to `CITATION.cff` and logged changes in `CHANGELOG.md`.

### Automation
Closes #789

### PR-Assignee

### Reviewer

### DIFF
--- a/open_mastr/xml_download/utils_write_to_database.py
+++ b/open_mastr/xml_download/utils_write_to_database.py
@@ -271,7 +271,8 @@ def interleave_files(threads_data: list):
     files_grouped_by_table = {}
 
     for item in threads_data:
-        table_name = item[2]
+        db_table = item[1]
+        table_name = db_table.name
         if table_name not in files_grouped_by_table:
             files_grouped_by_table[table_name] = []
         files_grouped_by_table[table_name].append(item)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
 
 
 ### Changed
+- Fix table name extraction in `interleave_files` to enable parallel worker interleaving and reduce SQLite lock contention
+  [#790](https://github.com/OpenEnergyPlatform/open-MaStR/pull/790)
 - Fix race condition in parallel bulk XML import that resulted in silent data loss
   [#765](https://github.com/OpenEnergyPlatform/open-MaStR/pull/765)
 - Add support for reading XSD files from both zipped / unzipped documentation sources
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -41,6 +41,9 @@ authors:
   - family-names: "Will"
     given-names: "Simon"
     alias: "@Simon-Will"
+  - family-names: "Rady"
+    given-names: "Hassan"
+    alias: "@hassanrady"
 identifiers:
   - type: swh
     value: 'swh:1:dir:28252b3ca57b56d22be851bad576b6748af2e171'
--- a/tests/xml_download/test_utils_write_to_database.py
+++ b/tests/xml_download/test_utils_write_to_database.py
@@ -9,7 +9,6 @@ import pytest
 from sqlalchemy import (
     Boolean,
     Column,
-    create_engine,
     Date,
     DateTime,
     Double,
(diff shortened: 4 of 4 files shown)


open_mastr/xml_download/utils_write_to_database.py
```diff
@@ -271,7 +271,8 @@ def interleave_files(threads_data: list):
     files_grouped_by_table = {}
 
     for item in threads_data:
-        table_name = item[2]
+        db_table = item[1]
+        table_name = db_table.name if hasattr(db_table, "name") else str(db_table)
```
Simon-Will, line 275: `db_table` should always be an SQLAlchemy Table object in my understanding. (I guess they can be pickled and shared with a different process that way.) And we rely on the `name` attribute being there throughout this whole file. So in my opinion, we can omit the `hasattr` guard here. Or am I overlooking something?

Simon-Will (changes requested):
Thanks for the quality PR, Hassan! It makes sense! I cannot fully test it right now because I'm on vacation and without a computer. @FlorianK13 or @kvnkrmr, maybe you could?

open_mastr/xml_download/utils_write_to_database.py
HassanRady: I included the `hasattr` guard was for backward compatibility with the existing legacy unit test `test_interleave_files()`, where the mock input tuples passed plain strings instead of `Table` objects:

```Python
("AnlagenEegBiomasse.xml", "anlageneegbiomasse", "anlageneegbiomasse")
```

FlorianK13: I have a lot to do right now :/ Maybe next week I find some time.

FlorianK13: Hi @HassanRady - is it a larger change needed to adapt the `test_interleave_files()`? If not I think I would prefer that way and not to carry both type cases in the actual code, meaning that db_table is of type sqlalchemy Table and not string. Thanks for your contribution anyway and sorry that it takes time to get a review :/

FlorianK13 (commented):
thanks for your pr, i added one comment.

FlorianK13:
I ran the download and it works well

open_mastr/xml_download/utils_write_to_database.py
HassanRady: If I understood correctly, since db_table will always be an SQLAlchemy Table object going forward, we should update the existing test (test_interleave_files) to use Table objects instead of plain strings, remove the hasattr guard from open_mastr/xml_download/utils_write_to_database.py, and remove the duplicate string-based test cases. Does that sound good?

FlorianK13: yes, i think that is good

FlorianK13:
@Simon-Will can you take a look if your requested changes are satisfied and if yes merge it?

Simon-Will:
Looks good, thanks a lot, @HassanRady!

Simon-Will (approved):

PR #793, opened 2026-09-04
Add script for testing before release

### LINKED ISSUE #771: Add release script that automates slow testing
A script that:
- downloads full mastr
- tests the final database. I would propose a lot of hard coded consistency checks for things that should be true for sure:
  - MaStR Nummer in all tables should match the regex
  - Bayern, Sachsen in Bundesland column in all tables
  - Power a float
  - ... and so on. Should be a large but maintainable number of these assumptions and they should be written in a way that they should stay true for a long period, ie no edge cases.

FlorianK13:
## Summary of the discussion

See #771 

Since this isn't code that goes into the package, I think a rough review is fine. First version of this script is LLM generated. I then ran the script and simplified it. Right now it prints one failure on Netze:Bundesland being Integer (ie not replaced by katalogwerte).

### PR-Assignee

### Reviewer


FlorianK13:
Whoever has time to review can do a review, but first take a look at my comments above :)

FlorianK13:
I only test sqlite for now - we could think about the same script running twice but once for postgres.

scripts/release_test.py
```diff
@@ -0,0 +1,347 @@
...
+    "Lokationtyp",
+    "Marktfunktion",
+    "Marktrolle",
+    "NetzbetreiberpruefungStatus",
+    "Nutzungsbereich",
+    "Personenart",
+    "Spannungsebene",
+    "Speicherart",
+    "Technologie",
+    "WindAnLandOderAufSee",
```
Simon-Will, line 88: We can add some more here that currently will not work. Cf. #769.

```suggestion
    "AnlageBetriebsstatus",
    "ArtDerSolaranlage",
    "ArtDerWasserkraftanlage",
    "Batterietechnologie",
    "Bundesland",
    "EinheitBetriebsstatus",
    "EinheitSystemstatus",
    "Einheittyp",
    "Einspeisungsart",
    "Energietraeger",
    "Ertuechtigungsart",
    "Hauptausrichtung",
    "Hauptbrennstoff",
    "Hersteller",
    "Land",
    "Lokationtyp",
    "Marktfunktion",
    "Marktgebiet",
    "Marktrolle",
    "NetzbetreiberpruefungStatus",
    "Nutzungsbereich",
    "Personenart",
    "Rechtsform",
    "Registergericht",
    "Spannungsebene",
    "Sparte",
    "Speicherart",
    "Technologie",
    "WeitereBrennstoffe",
    "WindAnLandOderAufSee",
```

scripts/release_test.py
```diff
@@ -0,0 +1,347 @@
...
+    "Breitengrad",
+    "Bruttoleistung",
+    "EinheitBetriebsstatus",
+    "Energietraeger",
+    "Gemeindeschluessel",
+    "Inbetriebnahmedatum",
+    "Laengengrad",
+    "Nettonennleistung",
+    "Registrierungsdatum",
+]
```
Simon-Will, line 102: Reasonable, I'd say. But how did you come up with exactly these columns?

scripts/release_test.py
```diff
@@ -0,0 +1,347 @@
...
+
+COLUMN_CHECKS = {
+    "*MastrNummer": "{col} REGEXP '^[A-Z]{3}[0-9]{12}([A-Z]{2})?"
+    "( *[,;] *[A-Z]{3}[0-9]{12})*$'",
+    "*MastrNummern": "{col} REGEXP '^[A-Z]{3}[0-9]{12}( *[,;] *[A-Z]{3}[0-9]{12})*$'",
+    "DatumDownload": "{col} REGEXP '^[0-9]{8}$'",
+}
+
+NOT_NULL_COLUMNS = ["DatenQuelle", "DatumDownload", "EinheitMastrNummer"]
```
Simon-Will: I think we can add some more here that should never be null because they are primary keys if they are present. I haven't checked this, though. Just an educated guess.

```suggestion
NOT_NULL_COLUMNS = [
    "DatenQuelle",
    "DatumDownload",
    "EinheitMastrNummer",
    "Id",
    "MastrNummer",
    "OpenMastrId",
]
```

Simon-Will (dismissed):
Thanks for the script, @FlorianK13! Looks very useful. I only have some emendations that we can also add later. So I have comments with suggestions, but I also approved. You could already merge if you like.

scripts/release_test.py
FlorianK13: Just some random columns - the idea is not to test all data that should be there but rather to test a few columns to make sure that columns are filled in general.

scripts/release_test.py
FlorianK13: We have a column OpenMastrId?

scripts/release_test.py
Simon-Will: Makes sense.

scripts/release_test.py
Simon-Will: Yes, we insert it for tables where we cannot identify the primary key. Currently, it is only used for the table `EinheitenAenderungNetzbetreiberzuordnungen`. See [this code here](https://github.com/OpenEnergyPlatform/open-MaStR/blob/2b44194374fe9ab72fac4750cc5e4743633ece0b/open_mastr/utils/sqlalchemy_tables.py#L30-L39).

PR #796, opened 2026-09-09
769: Add hard-coded catalog category columns

Simon-Will:
## Summary of the discussion

`bulk_cleansing=True` identifies MaStR catalog category fields from XSD enumeration restrictions. Some known category fields are declared as primitive types or unrestricted restrictions in the XSD, leaving their numeric IDs untranslated.

I'm adding an explicit table-scoped list of those XSD exceptions and classify them as catalog values during model generation. The existing bulk cleansing flow then replaces their IDs with values from `Katalogwerte.xml`.

In my pandas version, mixed columns of ints and comma-separated strings are inferred by pandas as `object`, not `string`. To correctly handle these columns, I modified the dtype check to also do the splitting of values for `object` columns. There's a regression test for that, too.

### Changed
- Translate catalog category IDs for fields that are missing XSD restrictions during bulk cleansing
  [#769](https://github.com/OpenEnergyPlatform/open-MaStR/issues/769)

### Automation
Closes #769

### PR-Assignee
- [x] 🐙 Follow the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)
- [x] 📝 Update the [CHANGELOG.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CHANGELOG.md)
- [ ] 📙 Update the documentation
  - Not applicable imo

### Reviewer
- [ ] 🐙 Follow the [Reviewer Guidelines](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md#40-let-someone-else-review-your-pr)
- [ ] 🐙 Provided feedback and show sufficient appreciation for the work done

PR #797, opened 2026-09-09
769 add manual category columns

FlorianK13:
New changes to #796 since it seems like I was not able to push to the forked repo of @Simon-Will 

EDIT: Seems like my changes were pushed to the branch of PR #796 but github was slow? I converted this back to draft and the discussion can be in the original PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and the versioning aims to respect [Semantic Versioning](http://semver.org/spec/
 
 
 ### Changed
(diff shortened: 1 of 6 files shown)


FlorianK13:
I added one handling of edge case: 
1. BNetzA decides that one of the columns is not an int anymore but the actual value, ie they do not write 345 anymore but "Bayern" in the Bundesland column.
2. The current code of #796 fails then because .astype("Int64") will fail
3. The changes here will ignore entries that have letters, ie it checks with the is_id variable that the entry only consists of digits or komma

PR #728, opened 2026-03-26
Zero fill specific date strings #719

pt-kkraemer:
## Summary of the discussion

When date-typos occur in years (0219 instead of 2019) and we do not pad these values when converting to string, the pandas `pd.read_sql` call cannot handle three-digit-years, it is simply not a correct isoformat, see #719.

### Updated
- Added `.str.zfill(10)` in `cast_date_columns_to_string`

## Workflow checklist

- [x] I have not yet checked other possible occurences in the code where we handle datetimes
- [x] I have to workaround the fact that `None.str.zfill(10)` does not work because `'NoneType' object has no attribute 'str'`
  - Fixed [here](https://github.com/OpenEnergyPlatform/open-MaStR/pull/728/changes/d14829447bbecb87d63d5a347386c9f717e52cd9)

### Automation
Closes #719

### PR-Assignee
- [x] 🐙 Follow the workflow in [CONTRIBUTING.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md)
- [x] 📝 Update the [CHANGELOG.md](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CHANGELOG.md)
- [x] 📙 Update the documentation

### Reviewer
- [ ] 🐙 Follow the [Reviewer Guidelines](https://github.com/OpenEnergyPlatform/open-MaStR/blob/production/CONTRIBUTING.md#40-let-someone-else-review-your-pr)
- [ ] 🐙 Provided feedback and show sufficient appreciation for the work done

Issue #719, opened 2026-01-27
ERROR Failed to process table market_actors: Invalid isoformat string: '201-10-01'


terminator-997:
The bulk download is failing with the below error:
ValueError: Invalid isoformat string: '201-10-01'
ERROR Failed to process table market_actors: Invalid isoformat string: '201-10-01'

I got this error on 22nd Jan and also today.

Can you please help with this?

FlorianK13:
This sounds like an error of the underlying data from BNetzA. Does the download crash after this error or does it continue?

terminator-997:
The download crashes after this error

pt-kkraemer:
I cannot reproduce the error, possibly because the file from today is correct. Can you provide us with the file from yesterday?

FlorianK13:
Anyway, if the download crashes due to a misformatted date string we should make it more robust, ie by parsing it:
* as a string (which could destroy the format of the whole column)
* as some random date (which looses the information from the given string)

huaqo:
I get a similar error, maybe this gives more information. My download keeps going after this:

[ERROR] Error processing file 'Marktrollen.xml': 'Index(['MastrNummer'], dtype='object')'

pt-kkraemer:
Hi @huaqo, that is a different kind of error which will probably be fixed in the next release, see #716.

terminator-997:
Hi @pt-kkraemer , I don't have the file as we are doing stateless run

terminator-997:
HI @pt-kkraemer , I did got the same error yesterday:
2026-01-28 15:27:55.995Z - ERROR Failed to process table market_actors: Invalid isoformat string: '201-10-01'

ramaroesilva:
Hi all. In a similar direction, I get a similar error - shown below - but only when running `db.to_csv(tables=['solar'])`

`ValueError: Invalid isoformat string: '221-12-21'`

SevKaloujski:
> Hi all. In a similar direction, I get a similar error - shown below - but only when running `db.to_csv(tables=['solar'])`
> 
> `ValueError: Invalid isoformat string: '221-12-21'`

Unfortunately I also have to report that it fails on db.to_csv() for solar with the same exact error, similar errors with market_players.
In addition, when downloading fails to download combustion_extended due to comma being present in the value

FlorianK13:
[...]

PR #801, opened 2026-09-15
Fix 800 handle malformed and out-of-range date and datetime data

### LINKED ISSUE #800: CSV export of "EinheitenSolar" crashes due to `ValueError: Invalid isoformat string: '205-12-06 00:00:00.000000'`
## Description of the issue

`db.to_csv()` raises:
```
Exporting the following database tables to CSV: EinheitenGeothermieGrubengasDruckentspannung, EinheitenStromSpeicher, EinheitenGasverbraucher, AnlagenEegSpeicher, AnlagenEegWasser, AnlagenStromSpeicher, EinheitenGasSpeicher, EinheitenSolar, EinheitenWasser, EinheitenKernkraft, Lokationen, AnlagenGasSpeicher, Marktakteure, EinheitenStromVerbraucher, EinheitenWind, EinheitenBiomasse, EinheitenGasErzeuger, Netze, MarktakteureUndRollen, AnlagenEegBiomasse, AnlagenKwk, AnlagenEegGeothermieGrubengasDruckentspannung, Bilanzierungsgebiete, EinheitenVerbrennung, Netzanschlusspunkte, AnlagenEegSolar, EinheitenGenehmigung, AnlagenEegWind
  File "/some/user/git-repos/open-MaStR/open-MaStR/main.py", line 61, in <module>
    db.to_csv()
    ~~~~~~~~~^^
  File "/some/user/git-repos/open-MaStR/open-MaStR/open_mastr/mastr.py", line 458, in to_csv
    for i, chunk in enumerate(
                    ~~~~~~~~~^
        pd.read_sql_table(requested_table_name, conn, chunksize=chunksize)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/some/user/.conda/envs/py310_open_mastr/lib/python3.14/site-packages/pandas/io/sql.py", line 1145, in _query_iterator
    data = result.fetchmany(chunksize)
  File "/some/user/.conda/envs/py310_open_mastr/lib/python3.14/site-packages/sqlalchemy/engine/result.py", line 1414, in fetchmany
    return self._manyrow_getter(self, size)
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/some/user/.conda/envs/py310_open_mastr/lib/python3.14/site-packages/sqlalchemy/engine/result.py", line 757, in manyrows
    rows = [make_row(row) for row in rows]
            ~~~~~~~~^^^^^
  File "lib/sqlalchemy/cyextension/resultproxy.pyx", line 22, in sqlalchemy.cyextension.resultproxy.BaseRow.__init__
  File "lib/sqlalchemy/cyextension/resultproxy.pyx", line 79, in sqlalchemy.cyextension.resultproxy._apply_processors
  File "lib/sqlalchemy/cyextension/processors.pyx", line 40, in sqlalchemy.cyextension.processors.str_to_datetime
ValueError: Invalid isoformat string: '205-12-06 00:00:00.000000'
```

Only 1 row in EinheitenSolar seems to be malformed:
Image

We saw something similar in the past in #719 .
While this seems to be a BNetzA data problem, I'd be good to have it fixed. I'll check again with new data next week.

I'll come up with a solution.

## Steps to Reproduce

Install conda env, current `develop` and run `main.py` (may use `to_csv("EinheitenSolar")`).

## Context and Environment
* Version used: `develop`, HEAD: 28b05a75378ace7fc765908d335d01d34e3212e0 , py3.14.7 with pandas 3.0.5
* Operating system: Ubuntu 26.04 LTS
* Environment setup and (python) version: conda

## Workflow checklist
- [x] I have checked [the documentation](https://open-mastr.readthedocs.io/en/latest/) and confirmed this issue is not already addressed there.

nesnoj:
tl;dr:
- CSV export fixed but discovered a general problem when parsing date columns which is not urgent and has to be discussed separately #802
- Full download and export works as expected
- All tests pass (some deprecation warnings which are unrelated to this PR)

## Summary of the discussion

`Mastr.to_csv` aborts with `ValueError: Invalid isoformat string: '205-12-06 00:00:00.000000'`. This was reported in #719 and supposedly fixed in #728, but that PR added `.str.zfill(10)` only to the `Date` branch of `cast_date_columns_to_string`. The value in the traceback comes from the `DateTime` branch (`%Y-%m-%d %H:%M:%S.%f`), which was never touched, so the bug survived. No test covered a year before 1000, which is why the gap went unnoticed.

The mechanism: `%Y` is not zero-padded for years before 1000 on Linux/glibc, so a MaStR typo such as `205` instead of `2005` is formatted as `205-12-06 …` instead of `0205-12-06 …`. A `DATETIME` column in SQLite has no dedicated storage class behind it, so the value is stored verbatim as TEXT. On read-back SQLAlchemy calls `datetime.fromisoformat`, which requires a four-digit year and raises. The import therefore succeeds and only the export fails. Whether the value reaches the database at all depends on the pandas version: with nanosecond-only datetimes `pd.to_datetime(errors="coerce")` turned the out-of-bounds year into `NaT`, while newer pandas infers a coarser resolution and keeps it. That is why the same defect resurfaced on the `DateTime` path.

This PR makes the export survive such a value instead of aborting on it. `Mastr.to_csv` no longer calls `pd.read_sql_table` directly. `_read_table_in_chunks` reflects the table, wraps `Date`/`DateTime` columns in `type_coerce(column, String)` so that SQLAlchemy's failing result processor never runs, and parses the raw strings with `pd.to_datetime(format="ISO8601", errors="coerce")`. A value that does not parse is reported with its table, column, count and up to three examples, and the cell is exported as empty.

`format="ISO8601"` is deliberate: without it pandas cannot infer a format from mixed good and bad values, falls back to dateutil, and dateutil parses `205-12-06` as year 205, so the cell would not be emptied. It accepts every valid variant (with or without microseconds, date-only, `NULL`) and also works on backends whose driver returns real `datetime` objects. The decision is made by whether the stored string parses, not by comparing its year against a threshold. CSV output for well-formed data was verified to be byte-identical to the previous read path.

## What this PR does not change

`open_mastr/xml_download/utils_write_to_database.py` stays untouched, and the zero-padding introduced by #728 stays in place. That padding is a separate problem, filed as a follow-up: it applies to `Date` columns only, and it decides what the year is, turning `205-12-06` into `0205-12-06` and `24-01-04` into `0024-01-04`, which are dates the register does not contain.

This is worth knowing when reading the export log, because it explains a result that is otherwise surprising. On a database written by the current code, a `Date` typo has been padded on import and parses fine, so it is exported unchanged and nothing is reported. Only a `DateTime` typo is still unreadable. A full export therefore produces a single warning, even though the underlying data holds many more malformed dates:

```
Table 'EinheitenSolar', column 'InbetriebnahmedatumAmAktuellenStandort': exporting
1 date value(s) without a four-digit year as empty values, e.g. '205-12-06 00:00:00.000000'
```

Fixing the padding would make the remaining 90 values surface here as well. That decision belongs to the follow-up issue, not to this PR, whose job is to stop the export from crashing.

## Verification

A full download of 2026-09-15 holds 40,042,729 rows across 28 tables. 91 date values have a year that is not four digits, 90 of them in `DATE` columns (`Marktakteure.Taetigkeitsbeginn` 58, `EinheitenSolar.DatumDesBetreiberwechsels` 25, `EinheitenStromSpeicher.DatumDesBetreiberwechsels` 5, `EinheitenVerbrennung.DatumDesBetreiberwechsels` 2) and 1 in a `DATETIME` column (`EinheitenSolar.InbetriebnahmedatumAmAktuellenStandort`, row `SEE938873874247`).

Reading that row through the old code path still raises `ValueError: Invalid isoformat string: '205-12-06 00:00:00.000000'`; through the new one it yields `NaT` and an empty cell. Exporting the affected tables from the real database completes, reports the values it empties, and leaves every well-formed date untouched.

### Changed
- Fix `Mastr.to_csv` aborting with "Invalid isoformat string" on date values whose year is not zero-padded to four digits. Those values are now exported as empty cells with a warning instead of failing the export
  [#801](https://github.com/OpenEnergyPlatform/open-MaStR/pull/801)

### Automation
Closes #800

### PR-Assignee

### Reviewer


nesnoj:
@FlorianK13 @Simon-Will could one of you give it a review?
Afterwards I can proceed with the linting stuff and we're heading towards v1.0.0 :rocket:

nesnoj:
@FlorianK13 BTW, I ran `release_test.py` (today's data), all tests passed except for column "Ertuechtigungsart" which seems to be not contained n the DB - maybe a format change?
Error:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) user-defined function raised exception                                                                                                                                           
[SQL: SELECT COUNT(*) FROM "Ertuechtigungen" WHERE "Ertuechtigungsart" IS NOT NULL AND NOT ("Ertuechtigungsart" REGEXP '^[^0-9]')]                                                                                                           
(Background on this error at: https://sqlalche.me/e/20/e3q8)
```

Simon-Will (approved):
Thanks for the contribution and the summary, @nesnoj! I think the CSV export fix is fine and I approve.

Frankly, I think the underlying problem is that we include these wrong date/datetime values in the database at all. But first making the CSV export robust against them is OK to me, as well.

I added a commit to pin down what happens with dates looking like `0205-01-01` to the new test. Hope that's alright.

Simon-Will:
> @FlorianK13 BTW, I ran `release_test.py` (today's data), all tests passed except for column "Ertuechtigungsart" which seems to be not contained n the DB - maybe a format change? Error:
> 
> ```
> sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) user-defined function raised exception                                                                                                                                           
> [SQL: SELECT COUNT(*) FROM "Ertuechtigungen" WHERE "Ertuechtigungsart" IS NOT NULL AND NOT ("Ertuechtigungsart" REGEXP '^[^0-9]')]                                                                                                           
> (Background on this error at: https://sqlalche.me/e/20/e3q8)
> ```

That's curious. I'm going to check if this happens for me later today.

Edit: I confirmed that and opened #804 for it.

nesnoj:
Thx for the review @Simon-Will !

